PR #5, opened 2025-07-31
update toml dependencies to workspace

nahimterrazas:
update toml dependencies to workspace
Api error type conversions under types


NicoMolinaOZ (approved):
LGTM! I believe this changes fix the issue with toml versions, thanks

pepebndc (approved):

PR #11, opened 2025-08-01
feat: Get order by id endpoint

NicoMolinaOZ:
- Adding get order by ID endpoint.


crates/solver-service/src/apis/order.rs
```diff
@@ -0,0 +1,140 @@
...
+) -> Result<OrderResponse, GetOrderError> {
+	// Extract data from the order's JSON data field
+	// This assumes the order.data contains the necessary fields for the API response
+	let input_amount = order
+		.data
+		.get("inputAmount")
+		.and_then(|v| serde_json::from_value::<AssetAmount>(v.clone()).ok())
+		.unwrap_or_else(|| AssetAmount {
+			asset: "0x0000000000000000000000000000000000000000".to_string(),
+			amount: alloy_primitives::U256::ZERO,
```
shahnami, line 89: I don't think we should be defaulting here. We need to gracefully handle them as errors. This goes for all the unwraps in this file..

crates/solver-service/src/apis/order.rs
```diff
@@ -0,0 +1,140 @@
...
+
+	let settlement_data = order
+		.data
+		.get("settlementData")
+		.cloned()
+		.unwrap_or_else(|| serde_json::json!({}));
+
+	// For now, assume all orders are pending unless we have more status tracking
+	// In a real implementation, you would query additional storage to determine the actual status
+	let status = DetailedIntentStatus::Pending;
```
shahnami, line 115: The main reason for this endpoint is status tracking, let's not default on one status

crates/solver-types/src/order.rs
```diff
@@ -78,3 +78,48 @@ pub struct FillProof {
...
+	pub output_amount: AssetAmount,
+	/// Settlement mechanism type
+	#[serde(rename = "settlementType")]
+	pub settlement_type: SettlementType,
+	/// Settlement-specific data
+	#[serde(rename = "settlementData")]
+	pub settlement_data: serde_json::Value,
+	/// Execution details if available
+	#[serde(rename = "executionDetails")]
+	pub execution_details: Option<serde_json::Value>,
```
shahnami, line 112: I wonder, what's the purpose of this field? Would it be transaction? Maybe we should call it that?

crates/solver-types/src/order.rs
```diff
@@ -78,3 +78,48 @@ pub struct FillProof {
...
+}
+
+/// Detailed intent status for API endpoints.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum DetailedIntentStatus {
+	Pending,
+	Executed,
+	Finalized,
+	Failed,
+}
```
shahnami, line 125: This is going to be mapped against SolverEvent? Maybe we should name it OrderStatus?

crates/solver-types/src/order.rs
```diff
@@ -107,17 +107,17 @@ pub struct OrderResponse {
 	/// Settlement-specific data
 	#[serde(rename = "settlementData")]
 	pub settlement_data: serde_json::Value,
-	/// Execution details if available
-	#[serde(rename = "executionDetails")]
-	pub execution_details: Option<serde_json::Value>,
+	/// Transaction details if order has been executed
```
shahnami, line 110: I'm guessing this would be the `fill` transaction?

crates/solver-service/src/apis/order.rs
```diff
@@ -0,0 +1,138 @@
...
+		.cloned()
+		.unwrap_or_else(|| serde_json::json!({}));
+
+	let status = order
+		.data
+		.get("status")
+		.and_then(|v| serde_json::from_value::<OrderStatus>(v.clone()).ok())
+		.ok_or_else(|| {
+			GetOrderError::Internal("Missing or invalid status field in order data".to_string())
+		})?;
```
shahnami, line 113: You will need to map this against `SolverEvent`. I'm pretty sure `OrderStatus` has a different format then the data you'll receive here.

crates/solver-types/src/order.rs
NicoMolinaOZ: I have renamed to fill_transaction

NicoMolinaOZ:
> Few more comments. How have you tested this?

Yes, here is the proof:
Screenshot 2025-08-04 at 3 10 37 PM
Screenshot 2025-08-04 at 3 12 08 PM

Screenshot 2025-08-04 at 3 11 09 PM

crates/solver-core/src/lib.rs
```diff
@@ -669,6 +669,50 @@ impl SolverEngine {
...
+					serde_json::json!(std::time::SystemTime::now()
+						.duration_since(std::time::UNIX_EPOCH)
+						.unwrap()
+						.as_secs()),
+				);
+				obj.insert(
+					"claim_tx_hash".to_string(),
+					serde_json::json!(hex::encode(&tx_hash.0)),
+				);
+			}
```
shahnami, line 707: Do we not have a set or update method on the storage interface?

crates/solver-core/src/lib.rs
```diff
@@ -669,6 +669,50 @@ impl SolverEngine {
...
+			.store(
+				"completed_orders",
+				&order_id,
+				&std::time::SystemTime::now()
+					.duration_since(std::time::UNIX_EPOCH)
+					.unwrap()
+					.as_secs(),
+			)
+			.await
+			.map_err(|e| SolverError::Service(e.to_string()))?;
```
shahnami, line 683: Not sure if we need this, what other tables exist already? E.g. "orders"? Then we can just retrieve by status or something?

crates/solver-core/src/lib.rs
```diff
@@ -669,6 +669,50 @@ impl SolverEngine {
...
+					serde_json::json!(hex::encode(&tx_hash.0)),
+				);
+			}
+
+			// Store the updated order
+			self.storage
+				.store("orders", &order_id, &order)
+				.await
+				.map_err(|e| SolverError::Service(e.to_string()))?;
+		}
```
shahnami, line 714: You should check if this updates automatically for any existing entries, or if it creates new entries all the time. Do we create the same table at other stages in the tx lifecycle?

crates/solver-service/src/apis/order.rs
```diff
@@ -0,0 +1,294 @@
...
+/// Validates the order ID format.
+fn validate_order_id(order_id: &str) -> Result<(), GetOrderError> {
+	// Remove UUID validation - accept any non-empty string as order ID
+	if order_id.is_empty() {
+		return Err(GetOrderError::InvalidId(
+			"Order ID cannot be empty".to_string(),
+		));
+	}
+
+	Ok(())
```
shahnami, line 71: Why did we remove this validation?

shahnami, line 71: Why did we remove this validation?

crates/solver-service/src/apis/order.rs
```diff
@@ -0,0 +1,294 @@
...
+	// Convert token bytes array to address
+	let token_array = output_token_bytes.as_array().ok_or_else(|| {
+		GetOrderError::Internal("Invalid token format - expected bytes array".to_string())
+	})?;
+
+	// Extract last 20 bytes (address part) from the 32-byte token field
+	let mut token_bytes = [0u8; 20];
+	for (i, byte_val) in token_array.iter().skip(12).take(20).enumerate() {
+		token_bytes[i] = byte_val.as_u64().unwrap_or(0) as u8;
+	}
```
shahnami, line 163: This seems like a workaround for something that should be easier. Why are we doing it this way?

crates/solver-service/src/apis/order.rs
```diff
@@ -0,0 +1,294 @@
...
+	let output_amount_u256 = output_amount_str
+		.parse::<alloy_primitives::U256>()
+		.map_err(|e| GetOrderError::Internal(format!("Invalid output amount: {}", e)))?;
+
+	let output_amount = AssetAmount {
+		asset: output_token,
+		amount: output_amount_u256,
+	};
+
+	// For EIP-7683, we can infer settlement type (default to Escrow for now)
```
shahnami, line 182: Add a TODO here

crates/solver-service/src/apis/order.rs
```diff
@@ -0,0 +1,294 @@
...
+		// Create a structured fill transaction object
+		Some(serde_json::json!({
+			"hash": format!("0x{}", alloy_primitives::hex::encode(&fill_tx_hash.0)),
+			"status": "confirmed", // or derive from actual status
+			"timestamp": chrono::Utc::now().timestamp() // you might want to store actual timestamp
+		}))
+	} else {
+		// Fall back to existing data in order.data if available
+		order.data.get("fillTransaction").cloned()
+	};
```
shahnami, line 211: Should we not just error?

crates/solver-service/src/apis/order.rs
```diff
@@ -0,0 +1,294 @@
...
+
+	let response = OrderResponse {
+		id: order.id,
+		status,
+		created_at: order.created_at,
+		last_updated: chrono::Utc::now().timestamp() as u64,
+		quote_id: None, // EIP-7683 orders don't have quote IDs by default
```
shahnami, line 218: That's not true.. ?

crates/solver-service/src/apis/order.rs
```diff
@@ -0,0 +1,294 @@
...
+	} else {
+		// Fall back to existing data in order.data if available
+		order.data.get("fillTransaction").cloned()
+	};
+
+	let response = OrderResponse {
+		id: order.id,
+		status,
+		created_at: order.created_at,
+		last_updated: chrono::Utc::now().timestamp() as u64,
```
shahnami, line 217: What does last updated represent? If it's the order's "Last updated" then it should not be "now".

crates/solver-service/src/apis/order.rs
```diff
@@ -0,0 +1,294 @@
...
+		quote_id: None, // EIP-7683 orders don't have quote IDs by default
+		input_amount,
+		output_amount,
+		settlement_type,
+		settlement_data,
+		fill_transaction,
+		error_details: order
+			.data
+			.get("errorDetails")
```
shahnami, line 226: Does this field exist in order data?

crates/solver-service/src/apis/order.rs
```diff
@@ -0,0 +1,294 @@
...
+	{
+		// Order has been executed (fill transaction submitted) but not yet finalized
+		return Ok(OrderStatus::Executed);
+	}
+
+	// Check if order has pending execution
+	if let Ok(_) = solver
+		.storage()
+		.retrieve::<(solver_types::Order, solver_types::ExecutionParams)>(
+			"pending_execution",
```
shahnami, line 271: I feel like we could have 1 orders table and track the status from there instead of 3-4 different tables..?

crates/solver-service/src/apis/order.rs
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	// Check for explicit failure indicators
+	if order.data.get("errorDetails").is_some() {
+		return Ok(OrderStatus::Failed);
+	}
+
+	// Fall back to checking stored status in order data
+	if let Some(stored_status) = order.data.get("status") {
+		println!("Stored status: {:?}", stored_status);
```
shahnami, line 286: Cruft

crates/solver-service/src/apis/order.rs
```diff
@@ -0,0 +1,294 @@
...
+			"pending_execution",
+			order_id,
+		)
+		.await
+	{
+		return Ok(OrderStatus::Pending);
+	}
+
+	// Check for explicit failure indicators
+	if order.data.get("errorDetails").is_some() {
```
shahnami, line 280: Please check if this field exists and where it gets populated?

crates/solver-service/src/apis/order.rs
```diff
@@ -0,0 +1,294 @@
...
+		.retrieve::<u64>("completed_orders", order_id)
+		.await
+	{
+		return Ok(OrderStatus::Finalized);
+	}
+
+	// Also check for completion status in order data (updated by core solver)
+	if let Some(status) = order.data.get("status") {
+		if status.as_str() == Some("Finalized") {
+			return Ok(OrderStatus::Finalized);
```
shahnami, line 253: Not sure if finalised should mean the solver claimed, it's ambiguous. Maybe it should be a separate status, and finalized would be after some block confirmations based on network speed? I'm open to other suggestions though.

crates/solver-service/src/apis/order.rs
NicoMolinaOZ: We are doing manual byte extraction because EIP-7683 stores addresses as 32-byte padded arrays, but  we need 20-byte hex strings. The "proper" fix would require type system changes across multiple components, so we are using this workaround for now.

crates/solver-service/src/apis/order.rs
NicoMolinaOZ: Not sure, should exist some scenario without fill information?

crates/solver-service/src/apis/order.rs
NicoMolinaOZ: Deleted from api response model, it does not exist for now.

crates/solver-service/src/apis/order.rs
NicoMolinaOZ: I have been asking the same, I am assuming when the solver claims the fee and it is confirmed. But not sure also..

crates/solver-service/src/apis/order.rs
NicoMolinaOZ: Good catch

crates/solver-core/src/lib.rs
shahnami: Yes let's add an update method please.

crates/solver-service/src/apis/order.rs
```diff
@@ -0,0 +1,215 @@
...
+		"signature": order.data.get("signature").cloned().unwrap_or_else(|| serde_json::json!(null)),
+		"nonce": order.data.get("nonce").cloned().unwrap_or_else(|| serde_json::json!(null)),
+		"expires": order.data.get("expires").cloned().unwrap_or_else(|| serde_json::json!(null))
+	});
+
+	// Try to retrieve fill transaction hash from storage
+	let fill_transaction = order.fill_tx_hash.as_ref().map(|fill_tx_hash| {
+		serde_json::json!({
+			"hash": format!("0x{}", alloy_primitives::hex::encode(&fill_tx_hash.0)),
+			"status": order.status,
```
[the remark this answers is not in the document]
shahnami: Alternatively, we can fetch status by transaction receipt hash, would be more direct, but require another RPC call.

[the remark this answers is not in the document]
NicoMolinaOZ: I have chosen the first option.

nahimterrazas (approved):
LGTM.

Minor comments.

Add documentation on _readme.md_ about the new endpoint.

crates/solver-core/src/lib.rs
```diff
@@ -526,23 +560,24 @@ impl SolverEngine {
 			}
 		};
 
-		// Retrieve pending execution details
-		let pending_data: Option<(Order, ExecutionParams)> = self
+		// Retrieve the full order with execution parameters
+		let order: Order = self
 			.storage
-			.retrieve("pending_execution", &order_id)
+			.retrieve("orders", &order_id)
```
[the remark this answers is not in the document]
NicoMolinaOZ: We don't have it at this moment, so it's worth creating a ticket to tackle this.

PR #13, opened 2025-08-02
fix: Scripts and implementations to match OIF contracts

shahnami:
This PR updates the OIF solver implementation to align with the latest OIF contracts structure and improves the reliability of the local development setup.

- Updated all solver implementations to match the new OIF contracts API (as of commit `f2a9e8`)
- Fixed EIP-7683 order handling in discovery, order processing, and settlement modules
- Updated intent structure parsing to handle the latest contract formats
- Gracefully handles errors: https://github.com/openintentsframework/oif-solvers/pull/14

**Demo Scripts Improvements**

- Fixed `setup_local_anvil.sh` reliability issues:
    - Added proper cleanup of log files and stale processes before startup
- Refactored `send_onchain_intent.sh` and `send_offchain_intent.sh` to use the updated contract interfaces
- Added `debug_transaction.sh`:
    - New utility script for debugging cross-chain transactions
- Added API server configuration to generated `demo.toml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,6 +316,11 @@ This script will:
 3. Wait for the solver to discover and fill the intent
 4. Show final balances demonstrating successful execution
 
+You can also check balances at any time using:
+```bash
+./scripts/demo/show_balances.sh
+```
+
 ### What the Demo Demonstrates
 
 1. **Intent Creation**: User deposits tokens into the InputSettler contract on the origin chain
--- a/scripts/demo/show_balances.sh
+++ b/scripts/demo/show_balances.sh
@@ -1,0 +1,110 @@
+#!/bin/bash
+# Show current token balances across both chains
+# Prerequisites: Run ./setup_local_anvil.sh first to generate config/demo.toml
+#
+# Usage:
+#   ./show_balances.sh
+
+set -e
+
+# Colors for output
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+NC='\033[0m' # No Color
+
+echo -e "${BLUE}📊 Current Token Balances${NC}"
+echo "========================="
+
+# Check required commands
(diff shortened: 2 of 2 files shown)


crates/solver-service/src/server.rs
shahnami, line 1: Changes are as a result of `cargo fmt`

scripts/demo/debug_transaction.sh
```diff
@@ -0,0 +1,635 @@
...
+        local contract_address=$2
+        local output_file=$3
+        
+        echo -e "\n${CYAN}📋 $contract_name ($contract_address)${NC}"
+        echo "----------------------------------------"
+        
+        # Extract errors
+        echo -e "${YELLOW}Errors:${NC}"
+        # Determine contracts directory relative to this script
+        CONTRACTS_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../oif-solvers/oif-contracts" && pwd)"
```
NicoMolinaOZ, line 75:
```suggestion
        CONTRACTS_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../oif-contracts" && pwd)"
```

NicoMolinaOZ (commented):
LGTM, I have tested the solution locally and is working.
Just one idea for future improvements and one minor change suggestion...

While the current send_offchain_intent.sh and onchain scripts get the job done, the section that hand-rolls EIP-712 hashes in Bash is becoming a maintenance headache (lots of cast calls, manual keccak,etc.). I’d like to suggest we should migrate only the EIP-712 signature generation into a small Rust helper using alloy crate.
I think it should be easier to maintain and for read the code.

nahimterrazas:
>  I’d like to suggest we should migrate only the EIP-712 signature generation into a small Rust helper using alloy crate. I think it should be easier to maintain and for read the code.

only migrate the EIP-712 signature and called from the bash, right?

PR #15, opened 2025-08-04
Remove redundant check balances script

shahnami:
- Remove redundant check balance script
- Update and format README
- Pin OIF contracts commit for demo scripts


nahimterrazas (approved):
LGTM

PR #60, opened 2025-08-08
chore: Prevent duplicate fills

shahnami:
- Adds additional storage read to check if an intent is already being processed
- Adds note to readme and demo scripts about test environment
- Adds millisecond scope to off-chain order NONCE to allow consecutive submissions without invalid order status errors.


NicoMolinaOZ (approved):
LGTM

nahimterrazas (approved):
LGTM

PR #69, opened 2025-08-11
refactor: Restructure quote processing into modular components

### LINKED ISSUE #68: Refactor Quote API: Separate Business Logic from HTTP Layer and Implement Settlement Strategy Decision Logic
Refactor - quote API implementation to properly separate concerns and implement the settlement strategy decision logic as per ERC-7683 and ERC-7930 standards.

nahimterrazas:
fix #68 

This code refactors api quote and makes it more modular.
- It uses storage for the quoting API endpoint.
- it separate the logic when is escrow and resourceLocks


shahnami (commented):
Looking good! Few comments, will do another look-over in a bit

PR #71, opened 2025-08-12
feat: Support multi-network/token configuration

shahnami:
This PR introduces a comprehensive Token Manager system that centralises token configuration and management across multiple networks. The implementation provides automatic token approval handling, balance monitoring, and a new API endpoint for querying supported tokens (useful for the aggregator).

**Features**

- Token Manager Module (`crates/solver-core/src/engine/token_manager.rs`)
  - Centralised token registry for multi-network support
  - Automatic `MAX_UINT256` approval management for configured tokens
  - Token balance monitoring and validation
  - Efficient approval state tracking to avoid redundant transactions
- Tokens API Endpoint (`crates/solver-service/src/apis/tokens.rs`)
  - New REST API endpoints for querying supported tokens
  - Network-specific token information retrieval
  - Complete token metadata including addresses, symbols, and decimals

**Refactored**

- Moved formatting and conversion utilities from `solver-core` to `solver-types` for better code organisation
  - `formatting.rs`: Token amount formatting with decimal support
  - `conversion.rs`: U256 conversion utilities
- Streamlined the engine context module by delegating token operations to the Token Manager
- Enhanced demo configuration with multi-network token support


nahimterrazas:
Please add ticket solved for this @shahnami

config/demo.toml
```diff
@@ -37,19 +62,17 @@ chain_id = 31338
 [discovery]
 [discovery.sources.onchain_eip7683]
 rpc_url = "http://localhost:8545"
-settler_addresses = ["0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0"]
+chain_id = 31337
```
nahimterrazas, line 65: Wouldn't be better directly reference from above:

`[networks.31337]
input_settler_address = "0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0"`

config/demo.toml
```diff
@@ -85,17 +108,21 @@ max_request_size = 1048576  # 1MB
 [contracts.origin]
 chain_id = 31337
```
nahimterrazas, line 109: same here

crates/solver-core/src/engine/token_manager.rs
```diff
@@ -0,0 +1,338 @@
...
+	///
+	/// Queries the current token balances for the solver's address on all
+	/// configured tokens and networks.
+	///
+	/// # Returns
+	///
+	/// Returns a HashMap mapping (chain_id, token_config) tuples to balance strings.
+	/// Balances are returned as decimal strings to avoid precision issues.
+	/// The token_config includes the token address, symbol, and decimals.
+	pub async fn check_balances(
```
nahimterrazas, line 239: Wouldn't it be good to add check_balance (one specific token address)?

shahnami: We could, but we don't need it for the moment. Anywhere else we do, we can call `delivery_service.get_balance`

config/demo.toml
shahnami: We could probably get rid of some of these values, they are/were used for demo scripts.

crates/solver-core/src/engine/token_manager.rs
nahimterrazas: I will need it for quote, that I was asking you :D

crates/solver-core/src/engine/token_manager.rs
```diff
@@ -0,0 +1,338 @@
...
+					.delivery
+					.get_allowance(
+						*chain_id,
+						&solver_address_str,
+						&hex::encode(&network.input_settler_address.0),
+						&hex::encode(&token.address.0),
+					)
+					.await?;
+
+				if current_allowance_input != max_uint256_str {
```
NicoMolinaOZ, line 129: I think a good practice here should be to set the allowance to zero value before increasing.
https://speedrunethereum.com/guides/erc20-approve-pattern

I’m also considering whether we should make an allowance for InputSettler. Is it required?

crates/solver-core/src/engine/token_manager.rs
```diff
@@ -0,0 +1,338 @@
...
+					.delivery
+					.get_allowance(
+						*chain_id,
+						&solver_address_str,
+						&hex::encode(&network.output_settler_address.0),
+						&hex::encode(&token.address.0),
+					)
+					.await?;
+
+				if current_allowance_output != max_uint256_str {
```
NicoMolinaOZ, line 155: Same here as above

config/demo.toml
shahnami: Looked into it, it's going to be a bigger refactor than originally anticipated. Will scope this in a new ticket.

crates/solver-service/src/apis/quote/generation.rs
```diff
@@ -15,289 +15,296 @@ use super::custody::{CustodyDecision, CustodyStrategy, EscrowKind, LockKind};
...
+			EscrowKind::Erc3009 => self.generate_erc3009_order(request, config),
+		}
+	}
+
+	/// Generate Permit2 SignatureTransfer order
+	fn generate_permit2_order(
+		&self,
+		request: &GetQuoteRequest,
+		config: &Config,
+	) -> Result<QuoteOrder, QuoteError> {
```
NicoMolinaOZ, line 141: I am thinking about this function, with the data created here, are we going to create an EIP-712 signature?

scripts/demo/send_offchain_intent.sh
```diff
@@ -5,7 +5,12 @@
 # NOTE: This script has been tested on macOS systems only.
 #
 # Prerequisites: Run ./setup_local_anvil.sh and start the solver service
-# Usage: ./send_offchain_intent.sh
+# Usage: 
+#   ./send_offchain_intent.sh [origin_token] [dest_token] [--direct|api_url]
+#   ./send_offchain_intent.sh                              # Use default TokenA
+#   ./send_offchain_intent.sh 0xABC... 0xDEF...          # Specific tokens
+#   ./send_offchain_intent.sh --direct                     # Use discovery service
+#   ./send_offchain_intent.sh 0xABC... 0xDEF... --direct  # Specific tokens + discovery
```
NicoMolinaOZ, line 13: I would add AMOUNT also as parameter, it could be useful for real world testing.
Another suggestion could be (feel free to discard): I think for most testnets chains the token contract address wont change, so we can create a dictionary, where can write: [origin_token]: USDT - [dest_token] : USDT
and we read internally the address, just an idea.

NicoMolinaOZ (commented):
Just some questions, feel free to discard if any of these make sense..

crates/solver-service/src/apis/quote/generation.rs
nahimterrazas: I dont think so, we are not able to do that, the user is the one who based on all this data, he creates the signature

crates/solver-core/src/engine/token_manager.rs
```diff
@@ -0,0 +1,338 @@
...
+//! The `TokenManager` acts as a central registry for all token-related operations
+//! in the solver system. It maintains knowledge of supported tokens across different
+//! blockchain networks and ensures that the solver has the necessary approvals to
+//! interact with these tokens through settler contracts.
+//!
+//! # Token Approvals
+//!
+//! The manager automatically sets MAX_UINT256 approvals for all configured tokens
+//! to their respective input and output settler contracts. This eliminates the need
+//! for per-transaction approvals and reduces gas costs during order execution.
```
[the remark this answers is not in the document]
shahnami: This should be fine, we imply trust if the solver-operator adds these networks and tokens in their config. There's more context on this [here](https://github.com/orgs/openintentsframework/projects/2/views/3?pane=issue&itemId=123183662&issue=openintentsframework%7Coif-solvers%7C47).

crates/solver-core/src/engine/token_manager.rs
shahnami: Thanks for the resource! However, I don't think it matters if we set it to max though. Additionally, we're not doing this on a per-intent basis, only at startup. We have the check in place so we don't make unnecessary calls as well.

scripts/demo/send_offchain_intent.sh
shahnami: That's a good idea. I'm going to be working on a bigger refactor for config files, which means I'll need to edit this file again. Will consider these changes then.

crates/solver-core/src/engine/token_manager.rs
NicoMolinaOZ: So you mean, are we executing the allowance only at startup? is a global config?

NicoMolinaOZ: Okay. Thanks for the response!

NicoMolinaOZ (approved):
LGTM! thanks!

PR #75, opened 2025-08-13
67 solver testing

NicoMolinaOZ:
Key changes:

 - Automated Testnet Deployment: Complete setup script (setup_testnet.sh) that deploys contracts to testnet chains available on testnet_chain.json file.
  - Batch Intent Testing: Sequential intent processor (batch_intents.sh) for testing multiple cross-chain intents
  - Multi-Network Support: Extended configuration to support additional testnet chains (Base Sepolia, Arbitrum Sepolia, OP Sepolia, Ethereum Sepolia)
  - Modular Configuration: Split testnet config into organized modules (networks.toml, api.toml)

### DIFF
--- a/crates/solver-config/src/builders/config.rs
+++ b/crates/solver-config/src/builders/config.rs
@@ -1,0 +1,134 @@
+//! Configuration builder for creating test and development configurations.
+//!
+//! This module provides utilities for constructing Config instances with
+//! sensible defaults, particularly useful for testing scenarios.
+
+use crate::{
+	AccountConfig, ApiConfig, Config, DeliveryConfig, DiscoveryConfig, OrderConfig,
+	SettlementConfig, SolverConfig, StorageConfig, StrategyConfig,
+};
+use std::collections::HashMap;
+
+/// Builder for creating `Config` instances with a fluent API.
+///
+/// Provides an easy way to create test configurations with sensible defaults.
+#[derive(Debug, Clone)]
+pub struct ConfigBuilder {
+	solver_id: String,
+	monitoring_timeout_minutes: u64,
+	storage_primary: String,
+	storage_cleanup_interval_seconds: u64,
+	min_confirmations: u64,
+	account_primary: String,
+	strategy_primary: String,
+	api: Option<ApiConfig>,
+}
+
+impl Default for ConfigBuilder {
+	fn default() -> Self {
+		Self::new()
+	}
+}
+
+impl ConfigBuilder {
+	/// Creates a new `ConfigBuilder` with default values suitable for testing.
+	pub fn new() -> Self {
+		Self {
+			solver_id: "test-solver".to_string(),
+			monitoring_timeout_minutes: 1,
+			storage_primary: "memory".to_string(),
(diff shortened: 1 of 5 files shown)


crates/solver-service/src/apis/order.rs
```diff
@@ -242,3 +242,295 @@ async fn convert_eip7683_order_to_response(
...
+			fn config_schema(&self) -> Box<dyn ConfigSchema>;
+			async fn cleanup_expired(&self) -> Result<usize, StorageError>;
+		}
+	}
+
+	const TEST_PK: &str = "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80";
+	const TEST_ADDR: &str = "0x1234567890123456789012345678901234567890";
+
+	fn test_cfg() -> Config {
+		Config {
```
shahnami, line 287: We could have a ConfigBuilder pattern for this with a default as well here

scripts/demo/batch_intents.sh
shahnami, line 1: Lets add this to another folder, maybe scripts/e2e, wdyt?

scripts/demo/test_intents.json
shahnami, line 1: Same here

scripts/demo/test_intents.json
```diff
@@ -0,0 +1,92 @@
...
+        "address": "0x5fd84259d66Cd46123540766Be93DFE6D43130D7",
+        "symbol": "USDC",
+        "decimals": 6
+      },
+      "amounts": {
+        "input": "1000000",
+        "output": "1000000",
+        "solver_fee": "0"
+      },
+      "api_mode": "default"
```
shahnami, line 23: What's this for?

scripts/demo/testnet_chains.json
shahnami, line 1: Same here

shahnami:
Looks good! Nice work @NicoMolinaOZ. There will be a few changes required once we merge the new oracle implementation. Also CI is failing..

scripts/demo/test_intents.json
NicoMolinaOZ: I have removed it, not necessary.

crates/solver-service/src/apis/order.rs
NicoMolinaOZ: I just pushed a possible solution, let's see..

shahnami: You can follow the pattern from monitors. E.g. `solver-config/builders/config.rs` imo.
I don't think you need the feature flag either, since we already import entire solver-config in solver-services anyway. Now you're importing twice as main and dev dependency? Is that pattern expected?

crates/solver-config/src/builders/config.rs
```diff
@@ -0,0 +1,134 @@
+//! Configuration builder for creating test and development configurations.
```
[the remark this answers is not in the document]
shahnami: Good point! It can and should be used for both. The current setup only defaults with test values so that should change..

PR #76, opened 2025-08-13
feat: Implement quote capability checks for destination balances

### LINKED ISSUE #23: Implement solver capability validation
- Check if the solver has the tokens required for the quote (on that chain) - requestedOutputs.asset
- Check if the solver has the balance required for the quote (on that chain) - requestedOutputs.amount
- Check if the solver has the gas required for the quote (ETH) - if calldata is empty, then gas transaction is always the same.
- Check if the solver has the time required for the quote (minValidUntil) - minValidUntil
- Check if the solver has the network required for the quote - if the chain is not in the whitelist, then the quote is not valid
- Check whitelist and blacklist of tokens - we need to match our list (Solver's list) with the availableInputs.asset
- Check for the oracle whitelist and blacklist - we need to match our list (Solver's list) with the requestedOutputs.asset
- determine if it is escrow or resourceLock.

nahimterrazas:
fixes #23 

feat: Implement quote capability checks for destination balances

- Added a new module `capability.rs` to handle runtime checks for sufficient token balances on destination chains before generating quotes.
- Integrated balance checks into the quote processing flow to ensure that the solver has enough liquidity for requested outputs.
- Enhanced the `QuoteValidator` with methods to validate supported networks and tokens, ensuring compliance with the solver's configuration.
- Updated the demo script to include scenarios for testing valid and invalid quote requests, including balance checks and unsupported tokens/chains.


crates/solver-service/src/apis/quote/validation.rs
```diff
@@ -3,12 +3,23 @@
...
 /// Handles validation of quote requests
 pub struct QuoteValidator;
 
+/// Representation of a token amount on a specific chain that the solver supports
+#[derive(Debug, Clone)]
+pub struct SupportedAsset {
+	pub chain_id: u64,
+	pub evm_address: AlloyAddress,
+	#[allow(dead_code)]
+	pub interop: InteropAddress,
```
shahnami, line 19: We should only need interop address right?

crates/solver-service/src/apis/quote/validation.rs
```diff
@@ -19,6 +30,117 @@ impl QuoteValidator {
...
+		for input in &request.available_inputs {
+			let chain_id = Self::chain_id_from_interop(&input.asset)?;
+			if let Ok(addr) = input.asset.ethereum_address() {
+				if let Some(net) = networks.get(&chain_id) {
+					if net
+						.tokens
+						.iter()
+						.any(|t| t.address.0.as_slice() == addr.as_slice())
+					{
+						found_supported = true;
```
shahnami, line 111: Can probably be replaced with tokenmanager.is_supported(...)

crates/solver-service/src/apis/quote/validation.rs
```diff
@@ -19,6 +30,117 @@ impl QuoteValidator {
...
+		// Check requested outputs
+		for output in &request.requested_outputs {
+			let chain_id = Self::chain_id_from_interop(&output.asset)?;
+			if let Ok(addr) = output.asset.ethereum_address() {
+				if let Some(net) = networks.get(&chain_id) {
+					if net
+						.tokens
+						.iter()
+						.any(|t| t.address.0.as_slice() == addr.as_slice())
+					{
```
shahnami, line 129: Same here

crates/solver-service/src/apis/quote/validation.rs
```diff
@@ -89,4 +211,105 @@ impl QuoteValidator {
...
+					)))
+				}
+			};
+
+			if let Some(net) = networks.get(&chain_id) {
+				if net
+					.tokens
+					.iter()
+					.any(|t| t.address.0.as_slice() == evm_addr.as_slice())
+				{
```
shahnami, line 249: I think the same here

crates/solver-service/src/apis/quote/validation.rs
```diff
@@ -89,4 +211,105 @@ impl QuoteValidator {
...
+
+			let supported = networks
+				.get(&chain_id)
+				.map(|net| {
+					net.tokens
+						.iter()
+						.any(|t| t.address.0.as_slice() == evm_addr.as_slice())
+				})
+				.unwrap_or(false);
+
```
shahnami, line 297: And here

crates/solver-types/src/validation.rs
shahnami, line 1: This is not relevant as part of this PR, but was missing some docs

crates/solver-service/src/apis/quote/validation.rs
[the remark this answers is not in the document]
nahimterrazas: I dont think so 🤔, but worth having a deep look. 

Also the test scenarios.

PR #96, opened 2025-08-21
refactor: Oracle implementations

shahnami:
Refactors the oracle architecture to support multiple oracle implementations per chain with configurable selection strategies. Adds route validation to ensure orders only flow between compatible oracle pairs, preventing settlement failures due to oracle mismatches.

- **Multi-Oracle Support:** Updated the system to support multiple oracles per chain instead of just one
- **Oracle Routes Validation:** Orders now validate that input and output oracles are compatible
- **Oracle Selection Strategies:** Added support for `First`, `RoundRobin`, and `Random` oracle selection

**Settlement Module:**
- Added `OracleConfig` with support for multiple oracles per chain
- Implemented oracle selection strategies
- Added route validation between input and output oracles

**Order Module:**
- Updated `Eip7683OrderImpl` to accept and validate oracle routes
- Added oracle compatibility validation in `validate_intent()`

**Configuration:**
- Updated config format to support multiple oracles and routes
- Fixed parsing issue with oracle_selection_strategy placement

**Fixes:**
- Removes index/lock files on cleanup
- Reduces storage expiry for demo configuration
- Fixes oracle selection
- Refactors settlement configuration
- Fixes trailing slashes in API


config/demo.toml
```diff
@@ -82,8 +82,21 @@ address = "0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0"
...
+# Oracle configuration with multiple oracle support
+[settlement.implementations.direct.oracles]
+# Input oracles (on origin chains)
+input = { 31337 = ["0xDc64a140Aa3E981100a9becA4E685f962f0cF6C9"], 31338 = ["0xDc64a140Aa3E981100a9becA4E685f962f0cF6C9"] }
+# Output oracles (on destination chains)
+output = { 31337 = ["0xDc64a140Aa3E981100a9becA4E685f962f0cF6C9"], 31338 = ["0xDc64a140Aa3E981100a9becA4E685f962f0cF6C9"] }
+
+# Valid routes: from origin chain -> to destination chains
+[settlement.implementations.direct.routes]
+31337 = [31338]  # Can go from origin to destination
```
pepebndc, line 98: This might be a problem when the solver starts scaling to many networks. I'm ok on keeping it like this for now, but I don't think it's scalable

shahnami: Could you clarify?
I imagine when the solver capabilities grow, we would add routes for supported networks like:

```
1: [2,3,4]
2: [1,5,10]
...
```

Are you saying this gets too complex or confusing or are you saying solvers would have a massive amount of routes they can support where this type of configuration stops making sense?

What type of configuration would you propose?

cc @luiz-lvj

luiz-lvj: imo the concept of "route" is important because the solver has to be aware of which oracles will be able to communicate between themselves, which imo translates into having these kind of "pairs" of input and output oracles. Besides that, it's not enough that two oracles are the same type (like "hyperlane" or "layerzero"), in general the oracles have to be aware of their counterpart, which makes the routes importants. 
Even in the context of the broadcaster for example, if I have a Base -> Arbitrum intent to be broadcasted, it's not enough to have a Broadcaster on Base and one Receiver on Arbitrum, one chain must have a copy the the other's prover, which has to be deployed by someone, so the route has to be created. 
This configuration solves that because it allows us to have a broadcaster on a L2 and it's important to know which other chains are able to receive the message later

cc @shahnami @pepebndc

config/demo.toml
```diff
@@ -82,8 +82,21 @@ address = "0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0"
 [settlement.implementations.direct]
 order = "eip7683"
 network_ids = [31337, 31338]
-oracle_addresses = { 31337 = "0xDc64a140Aa3E981100a9becA4E685f962f0cF6C9", 31338 = "0xDc64a140Aa3E981100a9becA4E685f962f0cF6C9" }
 dispute_period_seconds = 1
```
nahimterrazas, line 85: we need to work on 'dispute_period_seconds' this is more about optimistic oracles

config/demo.toml
```diff
@@ -82,8 +82,21 @@ address = "0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0"
...
+# Oracle selection strategy when multiple oracles are available (First, RoundRobin, Random)
```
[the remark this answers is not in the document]
shahnami: Yes, these will be in addition to user preference (security, speed, cost,...) provided or default to something based on maybe load (e.g. round robin).

config/demo.toml
```diff
@@ -14,16 +14,16 @@ monitoring_timeout_minutes = 5
...
 # Memory storage has no configuration
 
 [storage.implementations.file]
 storage_path = "./data/storage"
-ttl_orders = 0                  # Permanent
-ttl_intents = 86400             # 24 hours
-ttl_order_by_tx_hash = 86400    # 24 hours
+ttl_orders = 300                  # 5 minutes
+ttl_intents = 120                 # 2 minutes
+ttl_order_by_tx_hash = 300        # 5 minutes
```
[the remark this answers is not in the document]
shahnami: For demo, perhaps. This was just easier to test the cleanup logic with different values.

Issue #116, opened 2025-08-27
Get ETH - USD prices for quotation


nahimterrazas:
Get ETH - USD prices for quotation using
mock prices

PR #117, opened 2025-08-27
feat: introduce solver-pricing crate with mock pricing implementation…

nahimterrazas:
introduce solver-pricing crate with mock pricing implementation and integrate into solver-service

solves #116


crates/solver-pricing/src/implementations/mock.rs
```diff
@@ -0,0 +1,158 @@
...
+#[async_trait]
+impl PricingInterface for MockPricing {
+	fn config_schema(&self) -> Box<dyn ConfigSchema> {
+		Box::new(MockPricingSchema)
+	}
+
+	async fn get_asset_price(
+		&self,
+		asset: &str,
+		currency: &str,
```
shahnami, line 42: Can this also be ETH -> SOL or is this meant to always be priced against a fiat?

If the former, we should probably call this sourceAsset and destinationAsset or similar. If it's ALWAYS against fiat, maybe call it fiatCurrency?

crates/solver-service/src/apis/quote/cost/engine.rs
```diff
@@ -85,6 +72,7 @@ impl CostEngine {
 		solver: &SolverEngine,
 		config: &Config,
 		pricing: &PricingConfig,
+		pricing_service: &PricingService,
```
shahnami, line 75: We should remove pricingConfig here, since it should have been passed to PricingService already.

So anything requiring pricingConfig should be accessible internally by the service.

crates/solver-pricing/src/implementations/mock.rs
nahimterrazas: Yes, I changed the implementation a bit, now it is more flexible. Please check this out, and feel free to add more things

crates/solver-pricing/src/implementations/mock.rs
```diff
@@ -0,0 +1,281 @@
...
+		// Optional eth_price_usd field validation (legacy)
+		if let Some(price_value) = config.get("eth_price_usd") {
+			if price_value.as_str().is_none() {
+				return Err(ValidationError::TypeMismatch {
+					field: "eth_price_usd".to_string(),
+					expected: "string".to_string(),
+					actual: format!("{:?}", price_value),
+				});
+			}
+		}
```
shahnami, line 226: Probably don't need this if we support pair_prices field

crates/solver-pricing/src/implementations/mock.rs
```diff
@@ -0,0 +1,281 @@
...
+		&self,
+		wei_amount: &str,
+		currency: &str,
+	) -> Result<String, PricingError> {
+		let wei = U256::from_str_radix(wei_amount, 10)
+			.map_err(|e| PricingError::InvalidData(format!("Invalid wei amount: {}", e)))?;
+
+		// Convert wei to ETH (1 ETH = 10^18 wei)
+		let eth_decimals = U256::from(10_u64.pow(18));
+		let eth_amount_f64 = wei.as_limbs()[0] as f64 / (eth_decimals.as_limbs()[0] as f64);
```
shahnami, line 164: Try using Alloy parse_ether and format_ether helpers here?

crates/solver-pricing/src/implementations/mock.rs
```diff
@@ -0,0 +1,281 @@
...
+
+			if eth_price_f64 == 0.0 {
+				return Err(PricingError::InvalidData(
+					"ETH price cannot be zero".to_string(),
+				));
+			}
+
+			// Convert currency to ETH, then to wei
+			let eth_amount = currency_amount_f64 / eth_price_f64;
+			let wei_amount = eth_amount * 10_f64.powi(18);
```
shahnami, line 203: Same here

crates/solver-pricing/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+impl PricingConfig {
+	pub fn default_values() -> Self {
+		Self {
+			currency: "USD".to_string(),
+			commission_bps: 20,
+			gas_buffer_bps: 1000,
+			rate_buffer_bps: 14,
+			enable_live_gas_estimate: false,
```
shahnami, line 64: Do we need this?

crates/solver-service/src/apis/quote/mod.rs
```diff
@@ -127,9 +129,18 @@ pub async fn process_quote_request(
 		.get(strategy_name)
 		.unwrap_or(&default_table);
 	let pricing = cost::PricingConfig::from_table(strategy_table);
+
+	// Create a mock pricing service for now
+	// TODO: This should be properly injected or configured
```
shahnami, line 134: Is this something we can fix in this PR?

crates/solver-types/src/pricing.rs
```diff
@@ -0,0 +1,135 @@
...
+		Self {
+			base: base.to_uppercase(),
+			quote: quote.to_uppercase(),
+		}
+	}
+
+	/// Returns the pair as a string in format "BASE/QUOTE".
+	pub fn to_string(&self) -> String {
+		format!("{}/{}", self.base, self.quote)
+	}
```
shahnami, line 48: More idiomatic Rust way is to do something like this:

```

use std::fmt;

impl fmt::Display for TradingPair {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(f, "{}/{}", self.base, self.quote)
    }
}
```

crates/solver-types/src/pricing.rs
```diff
@@ -0,0 +1,135 @@
...
+	pub fn from_string(pair_str: &str) -> Result<Self, PricingError> {
+		let parts: Vec<&str> = pair_str.split('/').collect();
+		if parts.len() != 2 {
+			return Err(PricingError::InvalidData(format!(
+				"Invalid pair format: {}",
+				pair_str
+			)));
+		}
+		Ok(Self::new(parts[0], parts[1]))
+	}
```
shahnami, line 60:
```
use std::str::FromStr;

impl FromStr for TradingPair {
    type Err = PricingError;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        let parts: Vec<&str> = s.split('/').collect();
        if parts.len() != 2 {
            return Err(PricingError::InvalidPairFormat(s.to_string()));
        }
        Ok(Self::new(parts[0], parts[1]))
    }
}

```

crates/solver-types/src/pricing.rs
```diff
@@ -0,0 +1,135 @@
...
+	/// Gets the current price for an asset in the specified currency.
+	/// This is a convenience method that creates a TradingPair internally.
+	async fn get_asset_price(
+		&self,
+		asset: &str,
+		currency: &str,
+	) -> Result<AssetPrice, PricingError> {
+		let pair = TradingPair::new(asset, currency);
+		self.get_pair_price(&pair).await
+	}
```
shahnami, line 101: Potentially can be removed.

We can use get_pair_price with something just as ergonomic:

```
let pair: TradingPair = "ETH/USDC".parse()?;
let price = solver.get_pair_price(&pair).await?;
```

crates/solver-pricing/src/lib.rs
shahnami: Shouldn't we always be using "live gas estimates"? What's the alternative?

crates/solver-types/src/pricing.rs
```diff
@@ -0,0 +1,140 @@
...
+	/// Gets the current price for an asset in the specified currency.
+	/// This is a convenience method that creates a TradingPair internally.
+	async fn get_asset_price(
+		&self,
+		asset: &str,
+		currency: &str,
+	) -> Result<AssetPrice, PricingError> {
+		self.get_pair_price(&TradingPair::new(asset, currency))
+			.await
+	}
```
shahnami, line 106: I think you misunderstood, we can remove this entirely. It's a useless thin-wrapper.
Instead of calling `get_asset_price` anywhere, we can just call `self.get_pair_price(&TradingPair::new(asset, currency)).await` directly.

shahnami (approved):
LGTM, just the one comment.

PR #124, opened 2025-08-29
test: Unit tests solver account and delivery

NicoMolinaOZ:
# Summary

- Unit test cases for `solver-account` and `solver-delivery`
- Mock trait improvement inside `solver-service` tests
- Fix dependencies version

## Checklist


crates/solver-types/src/utils/builders/order.rs
```diff
@@ -0,0 +1,225 @@
...
+	prepare_tx_hash: Option<TransactionHash>,
+	fill_tx_hash: Option<TransactionHash>,
+	post_fill_tx_hash: Option<TransactionHash>,
+	pre_claim_tx_hash: Option<TransactionHash>,
+	claim_tx_hash: Option<TransactionHash>,
+	fill_proof: Option<FillProof>,
+}
+
+impl Default for OrderBuilder {
+	fn default() -> Self {
```
shahnami, line 31: If we're using this for production (not just testing), I wouldn't default to random values.. If we are using it _just_ for testing, maybe we need to switch them to a `tests/` folder to imply that?

shahnami: This goes for all builders btw

shahnami (approved):
Just a few comments to be tackled, I will already approve though.

crates/solver-types/src/utils/builders/networks.rs
```diff
@@ -468,6 +468,20 @@ impl NetworksConfigBuilder {
...
+	/// to be configured, allowing tests to create empty network configurations
+	/// to test error handling scenarios.
+	///
+	/// # Warning
+	///
+	/// This method is intended for testing purposes only. In production code,
+	/// use `build()` or `try_build()` which enforce proper validation.
+	pub fn build_empty_for_testing(self) -> NetworksConfig {
+		self.networks
+	}
```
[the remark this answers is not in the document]
NicoMolinaOZ: We have a validation to check if at least one network is configured. I can delete that validation to use NetworksConfigBuilder::new()

crates/solver-types/src/utils/builders/order.rs
NicoMolinaOZ: We are using only for testing atm, good suggestion, I can move all builder inside /test folder

nahimterrazas (approved):

Issue #104, opened 2025-08-26
use the quoteId generated to generate the intent


nahimterrazas:
On /orders API Post endpoint, we obtain from storage using orderId and execute the intent that the solver has committed to do.

follow [docs](https://docs.google.com/document/d/1qYyO3BQCeyrDn3LglGW7a65Q5VrTCtR9TlXe29f-e5g/edit?tab=t.0)

check deadlines.
check signature(s).

PR #136, opened 2025-09-02
104 quote intent generation

nahimterrazas:
This change introduce a connection between a quote and an intent

solves #104

### DIFF
--- a/crates/solver-types/src/api.rs
+++ b/crates/solver-types/src/api.rs
@@ -592,44 +592,6 @@ impl TryFrom<QuoteWithSignature<'_>> for serde_json::Value {
 		Ok(intent_request)
 	}
 }
-
-impl Quote {
-	/// Extract lock type from quote data with fallback logic
-	pub fn extract_lock_type(&self) -> String {
-		// First check the direct lock_type field on the quote
-		if !self.lock_type.is_empty() {
-			return self.lock_type.clone();
-		}
-
-		// Check if the quote contains lock type information in order data or witness data
-		if let Some(quote_order) = self.orders.first() {
-			if let Some(message_data) = quote_order.message.as_object() {
-				// Check for lock type in the main message data
-				if let Some(lock_type) = message_data.get("lockType").and_then(|v| v.as_str()) {
-					return lock_type.to_string();
-				}
-
-				// Check for lock type in EIP-712 data
-				if let Some(eip712_data) = message_data.get("eip712").and_then(|e| e.as_object()) {
-					if let Some(lock_type) = eip712_data.get("lockType").and_then(|v| v.as_str()) {
-						return lock_type.to_string();
-					}
-
-					// Check witness data for lock type
-					if let Some(witness) = eip712_data.get("witness").and_then(|w| w.as_object()) {
-						if let Some(lock_type) = witness.get("lockType").and_then(|v| v.as_str()) {
-							return lock_type.to_string();
-						}
-					}
-				}
-			}
-		}
-
-		// Default fallback
-		"permit2_escrow".to_string()
-	}
-}
-
 #[cfg(test)]
 mod tests {
 	use super::*;


shahnami (changes requested):
Thanks for starting work on this Nahim.

Can you help me understand why we aren't using the existing structure we have in server.rs instead?
 
There's an endpoint for `POST /quote` already. Moving this logic to the implementation level doesn't seem like a good idea to me. We need to consider separation of concerns. The discovery service shouldn't need to be aware that the concept of "quote" exists. A quote endpoint is specific to the solver, not to "discovering" intents.

Additionally it will complicate things around JWT management as well. Currently the expectation is that the intent route in discovery should be hosted internally, and made publicly accessible via the API service. So all auth management goes through one proxy.

I suggest we move the logic back to the API level (no need for `/accept` suffix either). Handle the conversion of quote -> intent and then forward the intent to offchain discovery API like we already do for POST intent routes.

Does that make sense?

shahnami:
> I know that this is not the best elegant way to do it, we need a refactor,. it deserves a more deep analisis, but that was the only way to implement without being too agressive @shahnami

Maybe i'm oversimplifying, but can't we just do something like:

1. Inside: `handle_committed_quote`
2. Retrieve order/intent from `StorageKey.Quotes` (accessible in `server.rs` -> `solver.storage()`)
3. Serialize to json payload
4. Forward to discovery service via `state.http_client.post(forward_url).json(&payload).send().await`

crates/solver-service/src/apis/quote/mod.rs
```diff
@@ -62,6 +62,7 @@
 //! - `InsufficientLiquidity`: Solver lacks required tokens
 //! - `UnsupportedChain`: Chain not configured
 //! - `Internal`: System errors
+// pub mod conversion; // Moved to discovery service
```
shahnami, line 65: I think we need to revert this one

crates/solver-types/src/api.rs
```diff
@@ -406,6 +408,232 @@ impl From<GetOrderError> for APIError {
 	}
 }
 
+impl Quote {
+	/// Convert quote and signature to IntentRequest format with encoded StandardOrder
+	#[cfg(feature = "abi")]
+	pub fn to_intent_request(
```
shahnami, line 414: I believe it would be more Rust idiomatic to use From<..>?

crates/solver-types/src/api.rs
```diff
@@ -406,6 +408,232 @@ impl From<GetOrderError> for APIError {
...
+				let chain_id = output_obj
+					.get("chainId")
+					.and_then(|c| c.as_u64())
+					.unwrap_or(0);
+				let amount_str = output_obj
+					.get("amount")
+					.and_then(|a| a.as_str())
+					.unwrap_or("0");
+				let token_str = output_obj.get("token").and_then(|t| t.as_str()).unwrap_or(
+					"0x0000000000000000000000000000000000000000000000000000000000000000",
```
shahnami, line 536: Lets be careful on the defaults here., we may not want to assume defaults

crates/solver-types/src/api.rs
```diff
@@ -406,6 +408,232 @@ impl From<GetOrderError> for APIError {
...
+					if let Some(lock_type) = eip712_data.get("lockType").and_then(|v| v.as_str()) {
+						return lock_type.to_string();
+					}
+
+					// Check witness data for lock type
+					if let Some(witness) = eip712_data.get("witness").and_then(|w| w.as_object()) {
+						if let Some(lock_type) = witness.get("lockType").and_then(|v| v.as_str()) {
+							return lock_type.to_string();
+						}
+					}
```
shahnami, line 627: This logic seems a bit wonky.. are there guarantees this will always hold true?

crates/solver-types/src/standards/eip7683.rs
```diff
@@ -505,3 +505,37 @@ mod tests {
 		assert_eq!(deserialized.context.len(), 500);
 	}
 }
+
+// Solidity struct definitions for ABI encoding with OIF contracts
+#[cfg(feature = "abi")]
```
shahnami, line 510: Nit;This is a great idea, but maybe we should scope it to standard instead of ABI?

crates/solver-types/src/api.rs
```diff
@@ -626,6 +854,7 @@ mod tests {
 			quote_id: "quote_123".to_string(),
 			provider: "test_solver".to_string(),
 			cost: None,
+			lock_type: "escrow".to_string(),
```
shahnami, line 857: `permit2_escrow` for consistency

crates/solver-discovery/src/implementations/offchain/_7683.rs
```diff
@@ -80,29 +83,7 @@ sol! {
...
-	struct SolMandateOutput {
-		bytes32 oracle;
-		bytes32 settler;
-		uint256 chainId;
-		bytes32 token;
-		uint256 amount;
-		bytes32 recipient;
-		bytes call;
-		bytes context;
+		function orderIdentifier(bytes calldata order) external view returns (bytes32);
```
shahnami, line 86: Did the signature change?

crates/solver-discovery/src/implementations/offchain/_7683.rs
```diff
@@ -704,17 +685,12 @@ impl Eip7683OffchainDiscovery {
...
 			LockType::ResourceLock => {
-				// Resource Lock (TheCompact) - use IInputSettlerCompact
-				let std_order = StandardOrder::abi_decode(order_bytes, true).map_err(|e| {
-					DiscoveryError::ParseError(format!("Failed to decode StandardOrder: {}", e))
-				})?;
+				// Resource Lock (TheCompact) - use IInputSettlerCompact with bytes
 				let compact = IInputSettlerCompact::new(settler_address, provider);
 				let resp = compact
-					.orderIdentifier(std_order)
+					.orderIdentifier(order_bytes.clone())
```
shahnami, line 693: Nice, that makes things cleaner!

crates/solver-service/Cargo.toml
```diff
@@ -9,6 +9,8 @@ name = "solver"
 path = "src/main.rs"
 
 [dependencies]
+alloy-sol-types = "0.8"
+hex = "0.4"
```
shahnami, line 13: Nit; sort

crates/solver-service/src/apis/quote/generation.rs
```diff
@@ -130,6 +137,7 @@ impl QuoteGenerator {
 			quote_id,
 			provider: "oif-solver".to_string(),
 			cost: None,
+			lock_type,
```
shahnami, line 140: Mind also pushing this change to oif-specs?

cc @pepebndc

shahnami (commented):
Looks much better! Thanks. Just a few more comments.

crates/solver-types/src/api.rs
```diff
@@ -406,6 +408,237 @@ impl From<GetOrderError> for APIError {
...
+			.get("expires")
+			.and_then(|e| e.as_u64())
+			.unwrap_or(quote_with_sig.quote.valid_until.unwrap_or(0)) as u32;
+		let fill_deadline = expires;
+
+		// Extract input oracle
+		let input_oracle_str = witness
+			.get("inputOracle")
+			.and_then(|o| o.as_str())
+			.unwrap_or("0x0000000000000000000000000000000000000000");
```
shahnami, line 488: Let's handle this also as error instead of fallback. It should actually fail anyway because we wouldn't support a route with ZERO address oracle..

crates/solver-types/src/api.rs
```diff
@@ -406,6 +408,237 @@ impl From<GetOrderError> for APIError {
...
+		// Parse outputs with proper helper function for hex parsing
+		fn parse_bytes32_from_hex(hex_str: &str) -> Result<[u8; 32], Box<dyn std::error::Error>> {
+			let hex_clean = hex_str.trim_start_matches("0x");
+			if hex_clean.len() != 64 {
+				return Err("Hex string must be exactly 64 characters (32 bytes)".into());
+			}
+			let mut bytes = [0u8; 32];
+			hex::decode_to_slice(hex_clean, &mut bytes)?;
+			Ok(bytes)
+		}
```
shahnami, line 536: Check our solver-types/src/utils?

crates/solver-types/src/standards/eip7683.rs
```diff
@@ -505,3 +505,49 @@ mod tests {
 		assert_eq!(deserialized.context.len(), 500);
 	}
 }
+
+// Solidity struct definitions for ABI encoding with OIF contracts
+#[cfg(feature = "standard")]
+pub mod standard {
```
shahnami, line 511: Probably rename to "interfaces" or something?

crates/solver-service/Cargo.toml
```diff
@@ -31,7 +33,7 @@ solver-order = { path = "../solver-order" }
 solver-pricing = { path = "../solver-pricing" }
 solver-settlement = { path = "../solver-settlement" }
 solver-storage = { path = "../solver-storage" }
-solver-types = { path = "../solver-types" }
+solver-types = { path = "../solver-types", features = ["standard"] }
```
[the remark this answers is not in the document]
nahimterrazas: Yeah but I dont think is eip7683, perhaps oif-interfaces /oif-contracts ?

crates/solver-types/src/standards/eip7683.rs
nahimterrazas: yes, I would prefer abi though haha

crates/solver-service/src/apis/quote/generation.rs
shahnami: I'm not sure if it needs to be different from what we store locally just yet.. how much do we differ from the current response (solver local vs solver to aggregator/user)

shahnami (approved):

crates/solver-types/src/api.rs
```diff
@@ -406,6 +408,237 @@ impl From<GetOrderError> for APIError {
...
+						if let Some(lock_type) = witness.get("lockType").and_then(|v| v.as_str()) {
+							return lock_type.to_string();
+						}
+					}
+				}
+			}
+		}
+
+		// Default fallback
+		"permit2_escrow".to_string()
```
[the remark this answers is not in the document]
nahimterrazas: actually we dont need that function anymore

PR #138, opened 2025-09-02
feat: Configure flexible expiration time for quote generation

NicoMolinaOZ:
# Summary

- Configure flexible expiration time for quote generation

## Checklist

### DIFF
--- a/config/demo.toml
+++ b/config/demo.toml
@@ -82,6 +82,14 @@ primary = "mock"
 # Uses default ETH/USD price of 4615.16
 
 # ============================================================================
+# QUOTE CONFIGURATION
+# ============================================================================
+[quote]
+# Quote validity duration in seconds
+# Default is 20 seconds. Customize as needed:
+validity_seconds = 60  # 1 minute validity
+
+# ============================================================================
 # SETTLEMENT
 # ============================================================================
 [settlement]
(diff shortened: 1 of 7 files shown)


config/demo.toml
```diff
@@ -81,6 +81,14 @@ primary = "mock"
...
 # Uses default ETH/USD price of 4615.16
 
+# ============================================================================
+# QUOTE CONFIGURATION
+# ============================================================================
+[quote]
+# Quote validity duration in seconds
+# Default is 20 seconds. Customize as needed:
+validity_seconds = 60  # 1 minute validity
+
```
shahnami, line 91: I think we can add this under:
```
[api]
[api.quotes]
....
```

Or something similar, wdyt?

NicoMolinaOZ: Emm yes, probably it makes sense.

crates/solver-settlement/Cargo.toml
```diff
@@ -12,11 +12,15 @@ alloy-sol-types = "0.8"
 alloy-transport = "0.8"
 alloy-transport-http = "0.8"
 async-trait = "0.1"
+mockall = { version = "0.13", optional = true }
```
NicoMolinaOZ, line 15: Regarding this change, here we have some rust projects using the same pattern, just in case you want to see:
https://github.com/serde-rs/serde/blob/master/serde/Cargo.toml

https://github.com/tokio-rs/tokio/blob/master/tokio/Cargo.toml

https://github.com/seanmonstar/reqwest/blob/master/Cargo.toml

Issue #147, opened 2025-09-11
Bug: Demo script not working


zeljkoX:
I have tried to run `./oif-demo env up` as described in documentation but run into issues:

- `oif-solvers/scripts/demo/lib/ui.sh: line 46: declare: -g: invalid option`
- `/oif-solvers/scripts/demo/lib/config.sh: line 29: declare: -g: invalid option`

Seems like this is due to bash version.

In case we want to support older version syntax and avoid users having issues we should modify script.

My bash version:
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin24)
Copyright (C) 2007 Free Software Foundation, Inc.

PR #148, opened 2025-09-11
fix: Removing global vars from demo scripts to be compatible with all bash version

NicoMolinaOZ:
# Summary

  - ui.sh - Removed -g flags from declare statements
  - config.sh - Replaced associative arrays with helper functions
  - intents.sh - Replaced INTENT_STATUS associative array with regular variables
  - deployment.sh - Replaced DEPLOYED_CONTRACTS associative array with regular variables
  - api.sh - Replaced API_RESPONSE associative array and removed -g flags
  - quotes.sh - Replaced QUOTE_STATUS associative array with regular variables

If you want to test with native bash version on macOS (in case you already update to v5), you can run:

`bin/bash oif-demo env up`

Issue: https://github.com/openintentsframework/oif-solver/issues/147

## Checklist


nahimterrazas (approved):
LGTM

Issue #152, opened 2025-09-15
Complete cost estimation logic for direct intents


shahnami:
I believe it should be using in the future to compare against max_allowed_cost? 🤔

_Originally posted by @NicoMolinaOZ in https://github.com/openintentsframework/oif-solver/pull/146#discussion_r2345319346_

PR #155, opened 2025-09-15
feat: Minimum solver profitability

NicoMolinaOZ:
# Summary

This PR implements minimum solver profitability checking to ensure the solver only processes orders that meet a configured profit threshold. The key changes include:

  - Configuration: Added min_profitability_pct field to solver configuration to set minimum required profit margin percentage
  - Profitability Calculation: Implemented order profitability analysis that calculates profit margin as (Total Input - Total Output - Execution Costs) / Total Input * 100
  - Order Filtering: Orders with profit margins below the threshold are rejected with an InsufficientProfitability error
  - Token Normalization: Added proper token amount normalization using token decimals for accurate cross-token profitability comparisons
  - Adding min profit value to the Quote total cost estimation.
  
Issue: https://github.com/openintentsframework/oif-solver/issues/152

## Checklist


crates/solver-service/src/server.rs
```diff
@@ -280,11 +282,40 @@ async fn handle_order(
 
 	// Estimate costs
 	let cost_engine = CostEngine::new();
-	let _order_cost = cost_engine
+	let order_cost = cost_engine
 		.estimate_cost(&validated_order, &state.solver, &state.config)
 		.await;
 
-	// TODO: Do something with _order_cost
+	// Check profitability if cost estimation succeeded
```
shahnami, line 289: What do we do if cost estimation failed? Accept anyway? Should we error / reject in that case?

crates/solver-service/src/server.rs
```diff
@@ -507,3 +538,112 @@ async fn forward_to_discovery_service(
 		},
 	}
 }
+
+/// Calculates the profit margin percentage for an order.
+///
+/// The profit margin is calculated as:
+/// Profit = Total Input Amount - Total Output Amount - Execution Costs
```
shahnami, line 545: This would only work if we're swapping the same value tokens (USDC -> USDC). We'll need to involve solver-pricing to get USD equivalent of a token (e.g. ETH -> USDC) we need get ETH/USDC pair token price and THEN compare normalised amounts.

crates/solver-service/src/server.rs
NicoMolinaOZ: I think we should reject the intent if we cannot calculate the profit

crates/solver-service/src/server.rs
```diff
@@ -507,3 +543,125 @@ async fn forward_to_discovery_service(
...
+	token_decimals: u8,
+	pricing_service: &solver_pricing::PricingService,
+) -> Result<String, Box<dyn std::error::Error>> {
+	// First normalize the token amount by dividing by 10^decimals
+	let raw_amount_str = raw_amount.to_string();
+	let raw_amount_decimal = Decimal::from_str(&raw_amount_str)
+		.map_err(|e| format!("Failed to parse raw amount: {}", e))?;
+
+	let divisor = Decimal::new(10_i64.pow(token_decimals as u32), 0);
+	let normalized_amount = raw_amount_decimal / divisor;
```
shahnami, line 660: Can you clarify why we're normalising this way? Why we need this and what it's actually doing?
Why is it dividing by 10, I would assume this depends from token to token (and the decimals they use?)

crates/solver-service/src/server.rs
```diff
@@ -507,3 +543,125 @@ async fn forward_to_discovery_service(
...
+	// Parse the order data as Eip7683OrderData
+	let order_data: solver_types::Eip7683OrderData = serde_json::from_value(order.data.clone())
+		.map_err(|e| format!("Failed to parse order data: {}", e))?;
+
+	let token_manager = solver.token_manager();
+	let pricing_service = solver.pricing();
+
+	// Calculate total input amount in USD (sum of all inputs converted to USD)
+	let mut total_input_amount_usd = Decimal::ZERO;
+	for input in &order_data.inputs {
```
shahnami, line 569: This type of parsing is again very specific to 7683 which has the ("inputs") field. Other order types might not have this, or it may be called differently..

crates/solver-service/src/server.rs
```diff
@@ -507,3 +543,125 @@ async fn forward_to_discovery_service(
...
+	tracing::debug!(
+		"Profitability calculation: input=${} (USD), output=${} (USD), cost=${} (USD), profit=${} (USD), margin={}%",
+		total_input_amount_usd,
+		total_output_amount_usd,
+		execution_cost_usd,
+		profit_usd,
+		profit_margin_decimal
+	);
+
+	Ok(profit_margin_decimal)
```
shahnami, line 644: This may make sense, but could we not just return the percentage, or do we need the decimal for some reason?

crates/solver-service/src/server.rs
```diff
@@ -280,11 +282,45 @@ async fn handle_order(
...
+		Ok(estimation) => estimation,
+		Err(api_error) => {
+			tracing::warn!("Cost estimation failed: {}", api_error);
+			return api_error.into_response();
+		},
+	};
 
-	// TODO: Do something with _order_cost
+	// Check profitability now that we have a valid cost estimate
+	match calculate_order_profitability(&validated_order, &cost_estimate, &state.solver).await {
```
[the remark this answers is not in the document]
NicoMolinaOZ: I thought Nahim was checking that, but in case not, I can do, of course

config/testnet/gas.toml
```diff
@@ -0,0 +1,13 @@
+[gas]
+
+[gas.flows.compact_resource_lock]
+# Gas units captured by scripts/e2e/estimate_gas_compact.sh
+open = 0
+fill = 76068
+claim = 121995
+
+[gas.flows.permit2_escrow]
```
shahnami, line 9: We're missing 3009_escrow gas prices here. Would you mind adding them to both testnet and demo configs (make sure to update deployments.sh and setup_testnet.sh)? You can get the values from ./oif-demo intent test escrow 3009 A2B and checking the tx receipts with forge for gas units used I believe.

shahnami: Or add a script in e2e similar to the estimate scripts we already have..

config/demo/gas.toml
```diff
@@ -11,3 +11,9 @@ claim = 121995
 open = 143116
 fill = 76068 
 claim = 59953
+
+[gas.flows.eip3009_escrow]
+# Gas units captured by scripts/e2e/estimate_gas_eip3009_escrow.sh on local anvil
+open = 130254
+fill = 77298 
+claim = 21000
```
shahnami, line 19: This seems off, let's get more realistic estimates from local anvil simulation.
Also don't forget that these files are generated from `deployments.sh` and `setup_testnet.sh` so adding the changes here won't be permanent. You'll need to update it in the relevant generation scripts.

shahnami (approved):

crates/solver-pricing/src/implementations/coingecko.rs
```diff
@@ -44,6 +44,8 @@ pub struct CoinGeckoPricing {
 	cache_duration: u64,
 	/// Map of token symbols to CoinGecko IDs
 	token_id_map: HashMap<String, String>,
+	/// Custom fixed prices for tokens (useful for testing/demo)
+	custom_prices: HashMap<String, String>,
```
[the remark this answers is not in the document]
shahnami: Agree, it was added as optional for that reason, makes testing a lot easier..

crates/solver-service/src/server.rs
```diff
@@ -302,12 +305,68 @@ async fn handle_order(
...
+	{
+		Ok(margin) => margin,
+		Err(e) => {
+			tracing::warn!("Failed to calculate profitability: {}", e);
+			return APIError::InternalServerError {
+				error_type: ApiErrorType::InternalError,
+				message: format!("Failed to calculate profitability: {}", e),
+			}
+			.into_response();
+		},
```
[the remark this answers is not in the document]
shahnami: This will be cleaned in a follow up PR nico is working on now, moving the cost engine to solver-core cleans up this a bit

config/demo.toml
```diff
@@ -76,7 +77,7 @@ max_gas_price_gwei = 100
 # PRICING
 # ============================================================================
 [pricing]
-primary = "mock"
+primary = "coingecko"
```
nahimterrazas, line 80: if we are keeping this, we need to change the script that generates on:

scripts/demo/lib/deployment.sh

Image

README.md
```diff
@@ -235,6 +236,8 @@ You can also use a single configuration file. See `config/example.toml` for a co
 [solver]
 id = "oif-solver-local"
 monitoring_timeout_minutes = 5
+# Minimum profitability percentage required for order execution
+min_profitability_pct = 1.0
```
nahimterrazas, line 240: great!

nahimterrazas (approved):

PR #157, opened 2025-09-16
feat: Add CoinGecko price feed implementation

shahnami:
# Summary

## Checklist

### DIFF
--- a/crates/solver-pricing/src/implementations/coingecko.rs
+++ b/crates/solver-pricing/src/implementations/coingecko.rs
@@ -1,0 +1,548 @@
+//! CoinGecko pricing implementation for production use.
+//!
+//! This implementation provides real-time asset prices from CoinGecko API.
+//! Supports ETH and other major cryptocurrencies against USD.
+
+use alloy_primitives::utils::parse_ether;
+use async_trait::async_trait;
+use reqwest::{
+	header::{HeaderMap, HeaderValue},
+	Client,
+};
+use serde::Deserialize;
+use solver_types::utils::wei_string_to_eth_string;
+use solver_types::{
+	ConfigSchema, ImplementationRegistry, PricingError, PricingFactory, PricingInterface,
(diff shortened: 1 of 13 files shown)


nahimterrazas (approved):
LGTM

crates/solver-pricing/src/implementations/coingecko.rs
```diff
@@ -0,0 +1,550 @@
...
+	rate_limit_delay_ms: u64,
+	/// Last API call timestamp
+	last_api_call: Arc<RwLock<Option<u64>>>,
+}
+
+/// CoinGecko API response for simple price endpoint
+#[derive(Debug, Deserialize)]
+struct SimplePriceResponse {
+	#[serde(flatten)]
+	prices: HashMap<String, HashMap<String, f64>>,
```
NicoMolinaOZ, line 59: Should we use f64 or rust_decimal::Decimal here?

NicoMolinaOZ (approved):
LGTM! just one minor question..

crates/solver-pricing/src/implementations/coingecko.rs
```diff
@@ -0,0 +1,550 @@
...
+	PricingRegistry, TradingPair, ValidationError,
+};
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use tokio::sync::RwLock;
+use tracing::debug;
+
+/// Precision for price formatting (decimal places)
+const PRICE_PRECISION: usize = 8;
```
NicoMolinaOZ, line 25: Does Coingecko always return prices with 8 digits of precision?

crates/solver-pricing/src/implementations/coingecko.rs
shahnami: Actually since we're using it for calculations, we should probably use decimals.. you're right

PR #160, opened 2025-09-16
feat: Implement compact signature validation for ResourceLock orders

nahimterrazas:
# Summary

This PR introduces a validation of the signature before we fill the intent on the destination chain when it is a compact order.

solves #31


crates/solver-order/src/implementations/standards/compact_signature_validator.rs
```diff
@@ -0,0 +1,475 @@
...
+		Ok(ValidationResult {
+			valid: true,
+			order_id: self.compute_order_id(order),
+			witness_hash: self.compute_witness_hash(order)?,
+			recovered_signer: expected_signer,
+		})
+	}
+
+	/// Compute order ID using the same logic as the contract
+	fn compute_order_id(&self, order: &OifStandardOrder) -> FixedBytes<32> {
```
shahnami, line 319: Pretty sure this is going to be wrong, can you verify it calculates the same order ID? I would just use the orderIdentifier view function call.

crates/solver-core/src/builder/mod.rs
```diff
@@ -403,7 +404,12 @@ impl SolverBuilder {
 		let mut order_impls = HashMap::new();
 		for (name, config) in &self.config.order.implementations {
 			if let Some(factory) = factories.order_factories.get(name) {
-				match factory(config, &self.config.networks, &oracle_routes) {
+				match factory(
+					config,
+					&self.config.networks,
+					&oracle_routes,
+					delivery.clone(),
```
shahnami, line 411: Let's try to avoid this

crates/solver-order/src/implementations/standards/_7683.rs
```diff
@@ -135,6 +148,213 @@ impl Eip7683OrderImpl {
 			OrderError::ValidationFailed(format!("Failed to decode StandardOrder: {}", e))
 		})
 	}
+
+	/// Validate compact signature for ResourceLock orders
+	async fn validate_compact_signature(
```
shahnami, line 153: I'm thinking this could be a function on the `Impl Eip7683OrderData {... validate_signature(..)... }`?

crates/solver-order/src/implementations/standards/_7683.rs
```diff
@@ -135,6 +148,213 @@ impl Eip7683OrderImpl {
...
+			expires: order_data.expires,
+			fillDeadline: order_data.fill_deadline,
+			inputOracle: input_oracle_address,
+			inputs: order_data.inputs.clone(),
+			outputs,
+		})
+	}
+
+	/// Fetch domain separator from TheCompact contract dynamically
+	async fn fetch_domain_separator(
```
shahnami, line 307: Let's make this `generate_domain_separator_transaction` and pass it to the solver-service layer above?

crates/solver-service/src/server.rs
```diff
@@ -385,6 +392,86 @@ fn create_intent_from_payload(payload: Value) -> Result<IntentRequest, APIError>
 	})
 }
 
+/// Validates EIP-712 signature using TheCompact protocol.
+async fn validate_signature(intent: &IntentRequest, state: &AppState) -> Result<(), APIError> {
```
nahimterrazas, line 396: wonder to move this inside 712? or other place that is not server.rs

shahnami: You have the right idea, but you see how you're only decoding OifStandardOrder? IntentRequest provides you an agnostic order type, but here we assume it will be Oif standard.. it needs to be agnostic. Similar to our validate_order. We do this at the implementation level. Maybe this indicates we should have something similar for validate_signature?

crates/solver-order/Cargo.toml
```diff
@@ -11,8 +11,10 @@ alloy-sol-types = "0.8"
 async-trait = "0.1"
 chrono = { version = "0.4", features = ["serde"] }
 hex = "0.4"
+secp256k1 = { version = "0.30", features = ["recovery"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+solver-delivery = { path = "../solver-delivery" }
```
shahnami, line 17: Let's remove this and clean up all the unnecessary files and code

crates/solver-order/src/implementations/standards/compact_signature_validator.rs
nahimterrazas: Since the calculation to validate the signature involves the orderId, naturally, it's the same.

So I think we can use this, it's working now

crates/solver-order/src/implementations/standards/_7683.rs
nahimterrazas: refactored

crates/solver-order/src/implementations/standards/_7683.rs
```diff
@@ -111,6 +110,7 @@ impl Eip7683OrderImpl {
 	///
 	/// * `networks` - Networks configuration with settler addresses
 	/// * `oracle_routes` - Oracle routes for validation
+	/// * `delivery` - Delivery service for blockchain interactions
```
shahnami, line 113:
```suggestion
```

crates/solver-order/src/implementations/standards/_7683.rs
```diff
@@ -465,6 +465,8 @@ impl OrderInterface for Eip7683OrderImpl {
 	/// Creates a transaction that calls the destination settler's `fill()` function
 	/// with the appropriate order data and solver information.
 	///
+	/// For ResourceLock orders, validates the signature before generating the fill transaction.
+	///
```
shahnami, line 469: Is this still true?

crates/solver-order/src/implementations/standards/_7683.rs
```diff
@@ -481,6 +483,7 @@ impl OrderInterface for Eip7683OrderImpl {
 	/// - Order is a same-chain order (not supported)
 	/// - No output exists for the destination chain
 	/// - Address parsing fails
+	/// - Signature validation fails for ResourceLock orders
```
shahnami, line 486: Let's verify comments are still valid after refactor

crates/solver-order/src/implementations/standards/_7683.rs
```diff
@@ -1011,10 +1016,9 @@ mod tests {
 
 		let result = Eip7683OrderImpl::new(networks, oracle_routes);
 		assert!(result.is_err());
-		assert!(result
-			.unwrap_err()
-			.to_string()
-			.contains("At least 2 networks"));
+		if let Err(e) = result {
+			assert!(e.to_string().contains("At least 2 networks"));
+		}
```
shahnami, line 1021: Why was this assertion changed?

crates/solver-order/src/implementations/standards/_7683.rs
```diff
@@ -1325,33 +1329,4 @@ mod tests {
 		assert!(LockType::Permit2Escrow.is_escrow());
 		assert!(!LockType::ResourceLock.is_escrow());
 	}
-
-	#[tokio::test]
-	async fn test_generate_fill_transaction_no_cross_chain_output() {
```
shahnami, line 1330: Same here, why was this test removed?

crates/solver-order/Cargo.toml
```diff
@@ -11,6 +11,7 @@ alloy-sol-types = "0.8"
 async-trait = "0.1"
 chrono = { version = "0.4", features = ["serde"] }
 hex = "0.4"
+secp256k1 = { version = "0.30", features = ["recovery"] }
```
shahnami, line 14: Still needed in this crate?

crates/solver-service/src/signature_validation.rs
shahnami, line 1: nit; rename to `signature_validator`?

PR #164, opened 2025-09-17
31 signature option 2

nahimterrazas:
# Summary

This introduces an abstract signature implementation

### DIFF
--- a/crates/solver-service/src/signature_validation.rs
+++ b/crates/solver-service/src/signature_validation.rs
@@ -21,6 +21,9 @@ use std::sync::Arc;
 /// Trait for validating signatures for specific order standards.
 #[async_trait]
 pub trait OrderSignatureValidator: Send + Sync {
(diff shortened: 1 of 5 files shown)


crates/solver-service/src/signature_validation.rs
```diff
@@ -0,0 +1,165 @@
...
+
+		validator
+			.validate_signature(intent, networks_config, delivery_service)
+			.await
+	}
+
+	/// Checks if signature validation is required for the given lock type.
+	pub fn requires_signature_validation(lock_type: &LockType) -> bool {
+		matches!(lock_type, LockType::ResourceLock)
+	}
```
shahnami, line 158: I think we should move this up into the implementation for Eip7683SignatureValidation as it may be different for other types potentially, or not?

shahnami (approved):

PR #167, opened 2025-09-18
162 support cost estimation for discovered intents

NicoMolinaOZ:
# Summary

## Checklist


crates/solver-core/src/engine/cost_profit.rs
```diff
@@ -0,0 +1,881 @@
...
+
+		let chain_params = ChainParams {
+			origin_chain_id,
+			dest_chain_id,
+		};
+
+		// Extract flow key (lock_type) for gas config lookup
+		let flow_key = Some(quote.lock_type.clone());
+
+		// Determine the standard - for quotes, we default to eip7683
```
shahnami, line 112:
```suggestion
		// TODO: pass standard as part of quote request
```

crates/solver-core/src/engine/cost_profit.rs
```diff
@@ -0,0 +1,881 @@
...
+
+		Ok(actual_profit_margin)
+	}
+
+	/// Validates cost estimation and profitability for an already-validated order from API requests.
+	///
+	/// This method combines cost estimation and profitability validation specifically for API-originated orders,
+	/// returning APIError types that can be properly handled by the HTTP layer.
+	/// For internally discovered intents, use the individual methods or IntentHandler directly.
+	pub async fn validate_order_profitability_for_api(
```
shahnami, line 374: Question: don't we tackle this as part of intent forwarding and eventual processing in intent.rs handler already?

shahnami (commented):
Looking good!

crates/solver-core/src/engine/cost_profit.rs
NicoMolinaOZ: So, yes, we do. However, on the main branch, we are returning an API error in case of a profit error. I thought about keeping the same functionality, as a user would be grateful for having that error beforehand. So with the processing inside `intent.rs` we cannot return API error

shahnami: So is it just for retaining "APIError" over some other error type? If that's the case, it's not worth doing cost calculations twice it can be an expensive computation..

NicoMolinaOZ: Emm okay, so we should only return 200 OK and skip the process after in case of profit margin error?

NicoMolinaOZ: I think this `source` field could be useful to decide if we should process the cost/profit 
https://github.com/openintentsframework/oif-solver/pull/165/files#diff-f05541fb614d61d934875ab5f916a585e72c7e5fc375ebfe89ec9abd4df809e0R192

PR #168, opened 2025-09-18
feat: Enhance EIP-712 support for BatchCompact orders

nahimterrazas:
# Summary

On going...

## Checklist

### DIFF
--- a/crates/solver-types/src/api.rs
+++ b/crates/solver-types/src/api.rs
@@ -336,58 +336,166 @@ impl IntentRequest {
 			.and_then(|e| e.as_object())
 			.ok_or("Missing 'eip712' object in message")?;
 
-		// Extract nonce
-		let nonce_str = eip712_data
-			.get("nonce")
-			.and_then(|n| n.as_str())
-			.ok_or("Missing nonce in EIP-712 data")?;
-		let nonce = U256::from_str_radix(nonce_str, 10)?;
+		// Determine order type based on primaryType
+		let primary_type = eip712_data
+			.get("primaryType")
+			.and_then(|p| p.as_str())
+			.unwrap_or("PermitBatchWitnessTransferFrom");
 
-		// Extract witness data
-		let witness = eip712_data
-			.get("witness")
-			.and_then(|w| w.as_object())
-			.ok_or("Missing 'witness' object in EIP-712 message")?;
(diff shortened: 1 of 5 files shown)


crates/solver-service/src/apis/quote/mod.rs
```diff
@@ -112,7 +112,8 @@ pub async fn process_quote_request(
 
 	// Generate quotes using the business logic layer
 	let settlement_service = solver.settlement();
-	let quote_generator = QuoteGenerator::new(settlement_service.clone());
+	let delivery_service = solver.delivery();
+	let quote_generator = QuoteGenerator::new(settlement_service.clone(), delivery_service.clone());
```
shahnami, line 116: If we're not using delivery_service anywhere else, could we just use solver.delivery() directly in the argument? Then we don't need to clone?

crates/solver-types/src/api.rs
```diff
@@ -303,34 +307,230 @@ impl TryFrom<(&Quote, &str, &str)> for IntentRequest {
...
+		// Detect order type from EIP-712 data
+		let (eip712_data, primary_type) = Self::extract_eip712_data(quote)?;
+
+		let sol_order = if primary_type == "BatchCompact" {
+			// Handle BatchCompact (ResourceLock) orders
+			Self::handle_batch_compact_order(quote, eip712_data)?
+		} else {
+			// Handle Permit2/EIP3009 orders using existing implementation
+			StandardOrder::try_from(quote)?
+		};
```
shahnami, line 513: What is handle_batch_compact_order doing structurally different than just try_from here?

crates/solver-types/src/standards/eip7683.rs
```diff
@@ -8,6 +8,10 @@ use alloy_primitives::U256;
 use serde::{Deserialize, Serialize};
 use std::str::FromStr;
 
+/// Type alias for EIP-712 data extraction result
+type Eip712ExtractionResult<'a> =
+	Result<(&'a serde_json::Map<String, serde_json::Value>, &'a str), Box<dyn std::error::Error>>;
+
```
shahnami, line 14: Do we have an eip712.rs file in solver-types that we can put this so its shared between here and api.rs?

PR #192, opened 2025-09-30
Improve cost estimation pt2

shahnami:
# Summary

## Checklist

### DIFF
--- a/crates/solver-service/src/apis/quote/mod.rs
+++ b/crates/solver-service/src/apis/quote/mod.rs
@@ -75,7 +75,7 @@ pub use validation::QuoteValidator;
 
 use solver_config::Config;
 use solver_core::SolverEngine;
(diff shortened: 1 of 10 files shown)


crates/solver-service/src/apis/quote/mod.rs
```diff
@@ -75,7 +75,10 @@ pub use validation::QuoteValidator;
```
NicoMolinaOZ, line 75: I dont know if we should use the `get_quote_by_id` method inside this file or keep the current `get_cost_context_by_quote_id` inside cost_profit.rs

NicoMolinaOZ (approved):
LGTM!

crates/solver-service/src/apis/quote/mod.rs
shahnami: Looking at it now, they are in different modules (solver-core vs solver-service) so let's keep the convenience methods..

nahimterrazas (approved):

PR #239, opened 2025-10-31
Validation Balance Compact

nahimterrazas:
This PR introduces a validation to check if the user has enough funds on the compact
 

## Summary by CodeRabbit

* **New Features**
  * Added order capacity validation with support for wallet balance and deposit verification across multiple lock types.
  * Introduced pluggable signature validation system with support for EIP-7683 standard orders.
  * Enhanced order validation workflow with pre-encoding capacity checks and improved error handling.


crates/solver-service/src/apis/order_validation.rs
```diff
@@ -0,0 +1,550 @@
...
+					token_id,
+					amount,
+				)
+				.await?;
+			}
+		},
+		_ => {
+			//  double check because when executing the openOrder will fail.
+			// We can reduce RPC Calls if we delete this validation
+			// But UX increase since will detect earlier if user has insufficient balance.
```
shahnami:
```suggestion
			// Defensive check: validates sufficient balance before calling `openFor`.
			// Trade-off: This adds an extra RPC call, but improves UX by failing fast
			// with a clear error rather than waiting for the `openFor` transaction to revert.
			// The `openFor` call would fail anyway with insufficient balance, so this
			// check is optional but recommended for better user experience.
```

shahnami (approved):
LGTM. Just renamed it to `order` and moved it into its own folder together with `signature_validators` -> `signature.rs`.

shahnami (approved):

PR #244, opened 2025-11-03
Fix hyperlane

nahimterrazas:
This PR basically adds 'from' to the transaction when calling estimate_gas_payment.

Also adds a delay to the RPC after the fill, this is because in a public node, sometimes it delays indexing, basically a race condition:

The Flow:
Fill transaction gets mined on chain (e.g., block 33216423)
Your solver confirms the Fill and immediately (0.2 seconds later) tries to:
Call quoteGasPayment() (read operation)
Submit a transaction to submit() (write operation)
The submit() function checks: hasAttested(payloads) which verifies if the Fill has been recorded in the OutputSettler contract
Public RPC Problem:
Public RPCs like https://sepolia.base.org use load balancing
They route your requests to different backend nodes
Node A might already have block 33216423 indexed → sees the Fill → returns true ✅
Node B might be 1-2 seconds behind → doesn't see the Fill yet → returns false ❌
Result: When you hit Node B, hasAttested() returns false, so submit() reverts with NotAllPayloadsValid

Also I added some logs.

## Summary by CodeRabbit

* **New Features**
  * Split post-fill monitoring into two explicit monitors: fill readiness (per-network RPC indexing delay) and claim readiness.
  * Gas-payment quoting now retries with exponential backoff.

* **Bug Fixes**
  * Improved settlement reliability by waiting for RPC indexing before post-fill progression and clearer claim monitoring triggers.
  * Enhanced error messages for transaction submission failures.

* **Refactor**
  * Unified and enriched transaction submission logging and sender-aware gas quoting.

* **Documentation**
  * Updated transaction state transitions and monitoring behavior.

* **Tests**
  * Test fixtures updated to include RPC indexing delay.

### DIFF
--- a/crates/solver-settlement/src/implementations/hyperlane.rs
+++ b/crates/solver-settlement/src/implementations/hyperlane.rs
@@ -1190,7 +1190,7 @@ impl SettlementInterface for HyperlaneSettlement {
 		// The contract will revert with "NotAllPayloadsValid" if the fill isn't visible yet
 		const MAX_ATTEMPTS: u32 = 8;
 		const INITIAL_DELAY_MS: u64 = 300;
-		
+
 		let mut gas_payment = None;
 		for attempt in 1..=MAX_ATTEMPTS {
 			match self
@@ -1236,9 +1236,11 @@ impl SettlementInterface for HyperlaneSettlement {
 				},
 			}
 		}
-		
+
 		let gas_payment = gas_payment.ok_or_else(|| {
-			SettlementError::ValidationFailed("Failed to get gas payment after all retries".to_string())
+			SettlementError::ValidationFailed(
+				"Failed to get gas payment after all retries".to_string(),
+			)
 		})?;
 
 		// Build submit call with correct payloads


crates/solver-settlement/src/implementations/hyperlane.rs
```diff
@@ -822,7 +822,6 @@ impl HyperlaneSettlement {
 		U256::from(base_gas + (payload_size * gas_per_byte) + buffer)
 	}
 
-	/// Estimate gas payment for a Hyperlane message
```
shahnami: Why remove this?

crates/solver-settlement/src/implementations/hyperlane.rs
```diff
@@ -1175,30 +1181,65 @@ impl SettlementInterface for HyperlaneSettlement {
...
-				origin_chain as u32,
-				recipient_oracle.clone(),
-				gas_limit,
-				vec![], // No custom metadata
-				output_settler.settler_address.clone(),
-				payloads.clone(),
-				order.solver_address.clone(), // Solver address for msg.sender
-			)
-			.await?;
+		// Smart retry: Attempt gas payment estimation with exponential backoff
```
NicoMolinaOZ: I think we already have logic to handle retries inside alloy.rs, but maybe does not fit for this specific case?

PR #248, opened 2025-11-11
Quote time config

nahimterrazas:
Enhance quote configuration with fill deadline and expiration settings

- Updated `api.toml` files for demo and testnet to include new parameters: `fill_deadline_seconds` (5 minutes) and `expires_seconds` (10 minutes).
- Modified `QuoteConfig` struct in `lib.rs` to add defaults for the new parameters.
- Updated quote generation logic in `generation.rs` to handle separate fill deadline and expiration timestamps.
- Added logging for deadline calculations in both `generation.rs` and `permit2.rs` to improve traceability of timing configurations.

## Summary by CodeRabbit

* **New Features**
  * Quotes now use two separate timing windows: a fill deadline (default 5 minutes) and an expiration (default 10 minutes), and generated orders/requests include both timestamps.
* **Chores**
  * Added new quote timing configuration keys and set default quote validity to 60 seconds.
  * Testnet API configuration now has API authentication disabled.


shahnami (approved):
LGTM. However, I think we should enforce these values on user request side to prevent us having to default to "solver" values. It seems odd that Solver is "deciding" on fill and claim deadlines. WDYT?

Other than that, I also agree with @pepebndc: we should add validation for on-chain intents as well to make sure we can actually fill and claim in a reasonable time.

PR #265, opened 2025-12-15
Fix postfill stuck

nahimterrazas:
# Summary

The root issue was in alloy’s [PendingTransactionBuilder::get_receipt()](https://github.com/alloy-rs/alloy/issues/389). It uses a header/subscription-based watcher that can miss the block where the tx was actually mined if the tx lands before the watcher is fully set up (race condition noted in alloy’s tracker). When that happens, get_receipt() keeps waiting for a “future” block containing the tx, even though the receipt already exists on-chain. So it hangs until timeout (8h in your config). That’s why you could fetch the receipt with cast but the monitor never finished.

We fixed it by replacing that watcher with a manual polling loop in crates/solver-delivery/src/implementations/evm/alloy.rs: directly call provider.get_transaction_receipt(tx_hash) every few seconds, check confirmations yourself, and bail on timeout. This removes the alloy watcher race and makes PostFill confirmations reliable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling and added tracing/logging for settlement callback failures and transaction confirmation flow.
  * Enhanced transaction monitoring with pre-mined receipt checks, stronger confirmation verification, clearer handling when transactions revert, and better distinction of monitoring failures.

* **Tests**
  * Expanded test coverage for monitoring behavior, provider/config validation, event/callback paths, multi-network/multi-signer scenarios, and error/confirmation cases.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/crates/solver-core/src/handlers/transaction.rs
+++ b/crates/solver-core/src/handlers/transaction.rs
@@ -107,7 +107,6 @@ impl TransactionHandler {
 					.handle_transaction_confirmed(&order, tx_type, &receipt)
 					.await
 					.map_err(|e: solver_settlement::SettlementError| {
-						tracing::error!("handle_transaction_confirmed failed: {:?}", e);
 						TransactionError::Service(format!("Settlement callback failed: {}", e))
 					})?;
 			}


crates/solver-core/src/handlers/transaction.rs
```diff
@@ -106,7 +106,8 @@ impl TransactionHandler {
 				settlement
 					.handle_transaction_confirmed(&order, tx_type, &receipt)
 					.await
-					.map_err(|e| {
+					.map_err(|e: solver_settlement::SettlementError| {
+						tracing::error!("handle_transaction_confirmed failed: {:?}", e);
```
shahnami: Do we still need this or was it just for debugging?

crates/solver-delivery/src/implementations/evm/alloy.rs
```diff
@@ -499,35 +501,100 @@ impl DeliveryInterface for AlloyDelivery {
...
+					"Transaction already confirmed"
+				);
+				return Ok(TransactionReceipt::from(&receipt));
+			}
+		} else {
+			tracing::error!(
+				tx_hash = %tx_hash,
+				chain_id = chain_id,
+				"Transaction reverted"
+			);
```
shahnami: Same here

PR #284, opened 2026-02-03
Kms implementation

nahimterrazas:
## Summary by CodeRabbit

* **New Features**
  * AWS KMS signing backend, unified signer abstraction, async account creation, and operator config now supports non-local account round‑trips.

* **Documentation**
  * Added AWS KMS Signing guide; SOLVER_PRIVATE_KEY marked optional and docs updated for multiple account backends.

* **Tests**
  * Many tests and helpers migrated to async; added signer and KMS-related tests.

* **Chores**
  * Rust toolchain bumped to 1.88.0 and AWS KMS support integrated.

* **Style**
  * Standardized error/log message formatting across the codebase.

solves #281

### DIFF
--- a/crates/solver-core/src/engine/token_manager.rs
+++ b/crates/solver-core/src/engine/token_manager.rs
@@ -617,7 +617,8 @@ mod tests {
 					.add_token(
 						TokenConfigBuilder::new()
 							.address(
-								parse_address("0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48").unwrap(),
+								parse_address("0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48")
+									.unwrap(),
 							)
 							.symbol("USDC")
 							.decimals(6)
@@ -631,7 +632,8 @@ mod tests {
 					.add_token(
 						TokenConfigBuilder::new()
 							.address(
-								parse_address("0xaf88d065e77c8cc2239327c5edb3a432268e5831").unwrap(),
+								parse_address("0xaf88d065e77c8cc2239327c5edb3a432268e5831")
+									.unwrap(),
 							)
 							.symbol("USDC")
 							.decimals(6)


crates/solver-account/src/lib.rs
```diff
@@ -8,11 +8,22 @@ use async_trait::async_trait;
...
+pub mod signer;
+
+/// Re-export AccountSigner for convenience
+pub use signer::AccountSigner;
+
 /// Re-export implementations
 pub mod implementations {
 	pub mod local;
+
+	#[cfg(feature = "kms")]
```
shahnami: Is there a specific reason to feature-gate this? I'm under the impression we'll probably use this as a default?

crates/solver-account/src/lib.rs
```diff
@@ -61,18 +72,38 @@ pub trait AccountInterface: Send + Sync {
...
 	///
-	/// This is required for all account implementations as it's used by
-	/// delivery implementations for transaction signing.
-	fn get_private_key(&self) -> SecretString;
+	/// # Panics
+	/// Panics if called on a KMS signer (use `signer()` instead).
+	///
+	/// # Deprecated
+	/// Use `signer()` instead - this method is not supported for KMS.
+	fn get_private_key(&self) -> SecretString {
```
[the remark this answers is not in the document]
shahnami: cc @pepebndc

crates/solver-core/src/builder/mod.rs
```diff
@@ -243,8 +251,8 @@ impl SolverBuilder {
 				match factory(
 					config,
 					&self.static_config.networks,
-					&default_private_key,
-					&network_private_keys,
+					&default_signer,
+					&network_signers,
```
shahnami: I like it, much cleaner!

crates/solver-delivery/src/implementations/evm/alloy.rs
```diff
@@ -605,25 +600,9 @@ pub fn create_http_delivery(
...
-					"Invalid private key format for network {}",
-					network_id
-				))
-			})
-		})?;
-		network_signers.insert(*network_id, signer);
-	}
+	// Clone the signers for use in the async block
+	let default_signer = default_signer.clone();
+	let network_signers = network_signers.clone();
```
shahnami: Makes me wonder why did we call with_exposed to extract private keys for local signers here before and why it's not needed anymore?

crates/solver-account/src/lib.rs
nahimterrazas: This is because we have these dependencies:

```
  alloy-signer-aws = { workspace = true, optional = true }
  aws-config = { workspace = true, optional = true }
  aws-sdk-kms = { workspace = true, optional = true }

```  
As an optional when we use KMS, we need to decide if we are going to keep it this way, or if we are going to support still private keys, get rid of this feature-gate ,and compile all together, whether or not we are not running KMS?

cc @pepebndc

crates/solver-account/src/implementations/local.rs
```diff
@@ -158,22 +162,26 @@ impl AccountInterface for LocalWallet {
...
-	LocalWalletSchema::validate_config(config)
-		.map_err(|e| AccountError::InvalidKey(format!("Invalid configuration: {}", e)))?;
-
-	let private_key = config
-		.get("private_key")
-		.and_then(|v| v.as_str())
-		.expect("private_key already validated");
-
-	Ok(Box::new(LocalWallet::new(private_key)?))
+pub fn create_account(config: &toml::Value) -> AccountFactoryFuture<'_> {
```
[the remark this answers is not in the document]
nahimterrazas: I created a new issue for solve this

crates/solver-account/src/lib.rs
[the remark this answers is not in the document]
shahnami: So is the signer approach not compatible with local signers?

crates/solver-delivery/src/implementations/evm/alloy.rs
shahnami: So this approach is no longer compatible with local signers?

PR #307, opened 2026-03-04
feat: Implement validation for cross-chain output oracles 

nahimterrazas:
## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened cross-chain validation to reject all-zero oracles and oracles with non-zero upper bytes; now requires exact destination address match and reports output index and destination chain in errors.
* **Tests**
  * Expanded test coverage for multi-output/multi-chain flows, adding cases for zero and dirty-upper-byte oracle values and index-aware assertions.


pepebndc (approved):

shahnami (approved):

pepebndc (changes requested):
Validated against the two output-oracle scan findings, **M-61** and **M-115**. M-61 is fully fixed; **M-115 is not addressed** — requesting changes (or confirmation that M-115 is tracked in a separate PR).

**M-61 (output oracle allows `0x00..00` / non-canonical `bytes32` padding) — ✅ fixed.** `validate_order` (`_7683.rs`, replacing the old `output.oracle != [0u8;32]` skip + `addresses_equal`) now rejects all-zero oracles for cross-chain outputs, rejects non-zero upper-12-bytes (`output_oracle[..12].iter().any(|&b| b != 0)`), and does a strict exact lower-20-byte match against a 20-byte supported oracle. Wired into the live admission path (`validate_and_create_order → validate_order`), with defense-in-depth in `generate_fill_transaction`/`generate_claim_transaction`. The dirty-upper-byte test is a genuine RED test (upper bytes 0xAA, lower 20 matching) that `addresses_equal` would have accepted. No regression (config oracles are always normalized to 20-byte addresses; zero-oracle orders were never settleable anyway).

**M-115 (admission doesn't validate order-bound output-oracle *membership*) — ❌ unaddressed.** This is a *separate* root cause in `crates/solver-settlement/src/admission.rs::implementation_supports_order`, which only checks oracle-list *existence* (`output_oracles.get(&destination).is_none_or(|o| o.is_empty())`) and never checks that the order-bound `output.oracle` is a *member* of that per-implementation allowlist. The PR touches only `_7683.rs`. Because `validate_order`'s `oracle_routes` are built from the **primary** settlement impl only (`build_oracle_routes`), an order pinning a **non-primary** `settlement_name` can still bind a non-member oracle, pass admission, get filled, and be unsettleable.

**Asks:**
1. Either implement M-115 here — parse the bound oracle (`parse_bound_output_oracle`) and enforce `output_oracles[destination].contains(...)` in `implementation_supports_order`, using the *pinned* settlement impl rather than primary-only routes — or confirm M-115 is tracked in a separate PR so this one can merge for M-61 alone.

**Minor:** M-61's `validate_order` enforcement uses primary-only `oracle_routes`; for non-primary pinned impls the strict comparison may consult the wrong allowlist source (overlaps with the M-115 fix).

PR #316, opened 2026-04-01
Rebalance implementation v2

nahimterrazas:
# Summary

## Checklist

## Summary by CodeRabbit

* **New Features**
  * LayerZero bridge for cross‑chain token rebalancing (composer & non‑composer flows)
  * Admin rebalance API: config, status, transfers, trigger, resolve, and threshold update endpoints
  * Deterministic threshold analysis and automatic rebalance triggering with cooldowns and delivery detection
  * Event log querying support for delivery service; EIP‑712 admin actions for rebalance

* **Bug Fixes / Reliability**
  * Robust transfer state machine, retries, timeouts, and persistent failure handling
  * Storage TTL/indexing improvements for terminal vs active transfers

* **Tests**
  * Extensive unit/integration tests covering bridge flows, monitor, storage, APIs, and thresholds


pepebndc:
🔴 **Blocker: `dest_chain` from request body is never validated**

📍 `crates/solver-service/src/apis/rebalance.rs` line 389

```rust
let (source_side, dest_side) = if request.source_chain == pair.chain_a.chain_id {
    (&pair.chain_a, &pair.chain_b)
} else if request.source_chain == pair.chain_b.chain_id {
    (&pair.chain_b, &pair.chain_a)
} else {
    return Err(axum::http::StatusCode::BAD_REQUEST);
};
```

`source_chain` is validated against the pair config, but `request.dest_chain` is accepted and used in the `BridgeRequest` without any validation. A caller could send `source_chain = chain_a` with `dest_chain = chain_a` (same chain) or an entirely wrong chain ID, and the bridge request would be constructed with an incorrect destination.

`dest_chain` should be derived from the pair config (i.e. `dest_side.chain_id`), not taken from the request body — or at minimum validated to equal the opposite side.

pepebndc:
🔴 **Blocker: 1:1 share-to-underlying assumption will silently redeem wrong amount**

📍 `crates/solver-bridge/src/monitor.rs` line 351

```rust
// Use the transfer amount as shares (for MVP, 1:1 with bridged amount)
let shares = U256::from_str_radix(&transfer.amount, 10).unwrap_or(U256::ZERO);
```

This uses the bridged amount directly as vault shares, but once the vault accumulates any yield the share-to-asset ratio diverges from 1:1 and this will silently redeem the wrong amount.

`previewRedeemCall` is already defined in `contracts.rs` but never used. Before production deployment, this should call `previewRedeem` to convert the bridged token amount to the correct number of shares (or use `previewWithdraw` to go from asset amount → shares).

pepebndc:
## Pending API Gaps

Comparing the product doc against the current implementation, these API-level items are missing:

1. **Monitor timing on status response** — `RebalanceStatusResponse` needs `last_check_at` and `next_check_at` fields. The monitor doesn't currently track or expose when it last ran or when the next tick is scheduled.

2. **Update config endpoint** — Need a `PUT /api/v1/admin/rebalance/config` (or similar) to modify targets, deviation bands, max amounts, and cooldown at runtime without redeploying.

3. **Toggle auto-rebalance** — The config has an `enabled` field but there's no endpoint to flip it. Needs at minimum a `POST .../enable` / `disable`, or a `PATCH` on the config.

4. **NeedsIntervention not documented** — The `/resolve` endpoint exists in code but the product doc doesn't mention it. The status response can return `NeedsIntervention(...)` but the doc only lists Pending, Relaying, Completed, Failed.

5. **Stable status/trigger serialization** — `status` and `trigger` fields in `BridgeOperationResponse` use `format!("{:?}", ...)` (Rust Debug format). `NeedsIntervention("reason")` is not machine-parseable. These need a proper `Display` impl or enum-to-string mapping with a separate `reason` field so the frontend can reliably match on states.

crates/solver-bridge/src/monitor.rs
```diff
@@ -73,22 +91,1125 @@ impl RebalanceMonitor {
...
+							)),
+						)
+						.await?;
+					continue;
+				},
+				_ => {},
+			}
+
+			// Poll bridge implementation for status update
+			let new_status = bridge_impl.check_status(&transfer).await?;
```
shahnami: `check_status` error here propagates via `?` and aborts the entire `for transfer in active` loop. A transient RPC failure for one transfer leaves all remaining transfers unprocessed for this tick.

I suggest to replace `?` with `match`/`if let Err` that logs the error and `continue`s to the next transfer.

crates/solver-bridge/src/monitor.rs
```diff
@@ -73,22 +91,1125 @@ impl RebalanceMonitor {
...
+							None,                                   // topic1: from (any)
+							Some(solver_types::H256(solver_topic)), // topic2: to = solver
+						],
+					);
+
+					match self.delivery.get_logs(transfer.dest_chain, filter).await {
+						Ok(logs) => {
+							let expected =
+								U256::from_str_radix(&transfer.amount, 10).unwrap_or(U256::ZERO);
+							let min_expected = expected * U256::from(95) / U256::from(100);
```
shahnami, line 448: The hardcoded 5%/105% delivery tolerance window could match unrelated `Transfer` events sent to the solver address with a similar amount. For a 1,000,000 transfer this accepts anything between 950,000 and 1,050,000.

Maybe you should make the tolerance configurable via `RebalanceConfig` or extract as a named constant with documentation explaining the trade-off.

crates/solver-service/src/apis/rebalance.rs
```diff
@@ -0,0 +1,1386 @@
...
+		.config_store
+		.get()
+		.await
+		.map_err(|e| AdminAuthError::Internal(format!("Config store error: {e}")))?;
+
+	let mut operator_config = versioned.data;
+
+	match operator_config.rebalance.as_mut() {
+		Some(rebalance) => {
+			rebalance.enabled = request.enabled;
```
shahnami, line 596: No bounds validation on `cooldown_seconds` or `max_pending_transfers`. Setting `cooldown_seconds` to 0 bypasses the cooldown safety guard entirely; setting `max_pending_transfers` to 0 blocks all auto-rebalances.

Maybe you should add range validation before persisting (e.g., `cooldown_seconds >= 30`, `max_pending_transfers` in `1..=10`).

crates/solver-bridge/src/monitor.rs
```diff
@@ -73,22 +91,1125 @@ impl RebalanceMonitor {
...
+						tracing::error!(pair = %pair.pair_id, "Auto-rebalance failed: {e}");
+					},
+				}
+			}
+		}
+
+		Ok(())
+	}
+
+	fn parse_address(addr: &str) -> Result<Address, crate::BridgeError> {
```
shahnami, line 817: nit: `parse_address` is duplicated identically here and in `layerzero/mod.rs:397`.

I suggest you extract to a shared utility in the crate root or `types.rs`.

crates/solver-bridge/src/monitor.rs
```diff
@@ -73,22 +91,1125 @@ impl RebalanceMonitor {
...
+		let mut topic = [0u8; 32];
+		topic[12..].copy_from_slice(&bytes);
+		H256(topic)
+	}
+
+	fn delivery_address(hex_addr: &str) -> DeliveryAddress {
+		DeliveryAddress(hex::decode(hex_addr.strip_prefix("0x").unwrap_or(hex_addr)).unwrap())
+	}
+
+	fn amount_data(amount: U256) -> Vec<u8> {
```
shahnami, line 969: nit: `amount_data` is just `amount.to_be_bytes::<32>().to_vec()` with extra steps.

Perhaps inline the expression.

shahnami (changes requested):
The persist-before-bridge pattern in rebalance_token is solid, and the NeedsIntervention fallback on save failures is a nice safety net. Shared threshold math between the monitor and status API prevents divergence, and test coverage across the state machine edge cases is thorough.

The main concern is the check_status error in the monitor tick loop aborting remaining transfers via early return, which could leave transfers stuck if one status check has a transient failure. A couple of API validation gaps around config updates could let operators set unsafe values (zero cooldown, zero max pending).

pepebndc (approved):

shahnami (approved):

PR #318, opened 2026-04-26
implementing exclusivity orders

nahimterrazas:
# Summary

## Checklist

## Release Notes

* **New Features**
  * Added exclusive-fill configuration for quotes supporting three modes: disabled, optional, or required
  * Quotes now embed solver-specific exclusivity context in order types
  * Configuration includes enforced time constraints with default and maximum duration settings

* **Tests**
  * Added tests validating exclusivity configuration, context generation, and constraint enforcement

### DIFF
--- a/crates/solver-config/src/lib.rs
+++ b/crates/solver-config/src/lib.rs
@@ -378,6 +378,64 @@ pub struct RebalancePairSideConfig {
 	pub oft_address: String,
 }
 
+/// Server policy for exclusive (limit-order) context on generated quotes.
+///
+/// When enabled, the solver populates `MandateOutput.context` with
+/// `0xe0 || bytes32(solver_address) || uint32(startTime)` so that the
+/// settlement contract reserves the fill for this solver until `startTime`.
+#[derive(Debug, Clone, Copy, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
+pub enum ExclusivityMode {
+	/// Never set exclusive context. Default. Back-compatible.
+	Disabled,
+	/// Set exclusive context only when the request opts in via metadata.
+	Optional,
+	/// Always set exclusive context.
+	Required,
+}
+
+impl Default for ExclusivityMode {
+	fn default() -> Self {
+		Self::Disabled
+	}
+}
+
+/// Configuration for exclusive-fill quotes.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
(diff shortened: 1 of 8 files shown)


pepebndc (changes requested):
Reviewed against **C-01** (Dutch-auction / exclusivity overpay). This is a useful quoter-side exclusivity *feature*, but it **does not remediate C-01**, and as branched it would **reopen the closed C-01 critical** — flagging before it leaves draft.

**What this PR does:** quote-generation side only — when configured, the solver embeds `0xe0 || bytes32(self) || uint32(startTime)` into its own quotes (naming *itself* `exclusiveFor`), via `ExclusivityConfig`/`ExclusivityMode` (default `Disabled`) and `encode_exclusive_context`, wired into the EIP-3009 / BatchCompact / Permit2 witness digests. Defaults are safe (disabled ⇒ bit-for-bit identical digests) and the encoding itself looks sound.

**Why it doesn't fix C-01:** the finding's root cause is entirely on the fill/admission/profitability side, and none of its three named functions are touched:
- (a) **No off-chain `_dutchAuctionPrice` replication** into the balance check + `validate_profitability` — `cost_profit.rs` still budgets on `output.amount` (the auction floor), the overpay root cause.
- (b) **No fill-side decode of a competitor's `exclusiveFor`/`startTime`** to avoid submitting inside another solver's window — the PR only *writes* `exclusiveFor=self`.
- (c) **No admission rejection of unsupported order-type bytes** — that guard is `validate_supported_output_context` from #371, which is **not on this branch** (forks from a pre-#371 commit).

**Blocking regression risk:** this branch also changes `StandardOrder::try_from` to *preserve* incoming output `context` bytes (previously emptied). Combined with the missing #371 admission guard, attacker-supplied Dutch/exclusive context would flow through to the unmodified `generate_fill_transaction` — i.e. the overpay/inventory-drain vector C-01 describes would be fully open.

**Asks before this can merge:**
1. Rebase onto #371 (the merged C-01 admission guard) and re-verify the interaction — the solver must not admit `0x01`/`0xe0`/`0xe1` it can't price.
2. If the intent is to actually *support* (not just reject) Dutch/exclusive orders, implement C-01 (a) and (b): off-chain decayed-price replication feeding the profitability gate, and competitor-window detection on the fill side.
3. Add a test exercising the *fill/profitability* path with a Dutch (`0x01`) / exclusive (`0xe1`) order — current tests only cover quote emission of the self-exclusivity marker.

Keeping it as draft is right. The quoter feature itself is fine; it just shouldn't be read as the C-01 remediation.

PR #319, opened 2026-04-26
Flow continuity hardening

nahimterrazas:
# Summary

## Checklist

## Summary by CodeRabbit

* **New Features**
  * On-chain nonce manager with resync/retry, ERC‑20 allowance precheck to skip unnecessary approves, and configurable tx‑confirmation polling with an explicit "indeterminate" monitoring outcome.

* **Bug Fixes**
  * Better recovery for missing redeems and in‑flight approve/crash‑window scenarios; native‑gas shortfalls now surface as NeedsIntervention with diagnostic details.

* **API**
  * Rebalance responses expose approve lifecycle fields and richer statusReason diagnostics.

* **Tests**
  * Expanded coverage for nonce logic, approve/approve-skip, redeem recovery paths, error mapping, and indeterminate monitoring.

* **Documentation**
  * Operator guidance for insufficient native gas and retry workflow.

### DIFF
--- a/crates/solver-bridge/src/monitor.rs
+++ b/crates/solver-bridge/src/monitor.rs
@@ -22,6 +22,12 @@ use tokio::sync::{watch, RwLock, Semaphore};
 /// Timeout for Submitted state before NeedsIntervention.
 const SUBMITTED_TIMEOUT_SECS: u64 = 30 * 60;
 
+/// Absolute upper bound on the approve phase. Catches a low-gas approve that
+/// stays in the mempool forever (tx_exists keeps returning true). After this,
+/// the only safe action is admin intervention — automatic fee bumping is
+/// deferred (requires Fix 3 / TxAttempt outbox).
+const APPROVE_PHASE_TIMEOUT_SECS: u64 = 60 * 60; // 1 hour
+
 /// Timeout for Relaying state before NeedsIntervention.
 const RELAYING_TIMEOUT_SECS: u64 = 30 * 60;
 
@@ -30,6 +36,10 @@ const PENDING_REDEMPTION_TIMEOUT_SECS: u64 = 24 * 3600;
 
 /// Max retries for PendingRedemption before NeedsIntervention.
 const MAX_REDEEM_RETRIES: u32 = 3;
+
+fn is_source_transaction_missing(reason: &str) -> bool {
+	reason.contains("Source transaction not found") || reason.contains("Source transaction missing")
+}
 
 /// Background rebalance monitor.
 /// Shared monitor timing state, readable by the admin status API.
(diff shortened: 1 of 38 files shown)


crates/solver-delivery/src/implementations/evm/nonce.rs
```diff
@@ -0,0 +1,246 @@
...
+	/// If the current cached value is already at or above `next`, no change.
+	/// Returns the cached value after the call.
+	pub fn set_next_nonce(&self, address: Address, next: u64) -> u64 {
+		let mut cache = self.cache.lock().expect("nonce cache mutex poisoned");
+		let entry = cache.entry(address).or_insert(next);
+		if *entry < next {
+			*entry = next;
+		}
+		*entry
+	}
```
luiz-lvj, line 57: why not fetch the catch every time? Is there a reason to keep the cache?

crates/solver-delivery/src/implementations/evm/nonce.rs
```diff
@@ -0,0 +1,246 @@
...
+
+	/// Take the next nonce and advance the counter.
+	/// Returns `None` if no nonce has been cached yet for this address —
+	/// callers should fetch chain-pending and call `set_next_nonce` first.
+	pub fn take_next(&self, address: Address) -> Option<u64> {
+		let mut cache = self.cache.lock().expect("nonce cache mutex poisoned");
+		let n = cache.get_mut(&address)?;
+		let taken = *n;
+		*n = taken + 1;
+		Some(taken)
```
luiz-lvj, line 96: what happens if the cache is n but the actual nonce is n + 10 for example? will this function be called in loop get 10 "nonce too low" errors?

crates/solver-bridge/src/monitor.rs
```diff
@@ -823,6 +1108,187 @@ impl RebalanceMonitor {
...
+		};
+
+		// CRASH-WINDOW GUARD: persist marker BEFORE the bridge_asset call.
+		transfer.bridge_submit_attempted = true;
+		self.bridge_service
+			.storage()
+			.save_transfer(transfer)
+			.await?;
+
+		let _permit = self.transaction_semaphore.acquire().await.map_err(|e| {
```
shahnami, line 1278: This semaphore acquire covers the entire `bridge_asset` call, including the confirmation polling loop (up to 12 × 5s = 60s in production). That means an approve-resume holds the sole permit for up to a minute, blocking any concurrent pair submission. The redeem path at line 557 uses a narrower pattern, it acquires the semaphore only around `deliver` and releases it before polling. Would it make sense to apply the same pattern here, or is serializing through confirmation intentional? Either way a short comment would help clarify.

shahnami (changes requested):
One critical issue before merging: the `resolve_transfer` "retry" arm resets `failure_count` and restores `status_before_intervention` but never clears `bridge_submit_attempted = false`. A transfer escalated to `NeedsIntervention` via the generic error path has `bridge_submit_attempted = true` and no `tx_hash`, after an admin retry the restored `Submitted` status immediately re-triggers the crash-window guard on the next tick, looping back to `NeedsIntervention`. 

I suggest to fix by adding `transfer.bridge_submit_attempted = false;` in the "retry" arm, since the operator's explicit action implies they've confirmed no deposit was broadcast.

Minor: `get_vault` in `layerzero/mod.rs:86` has `#[allow(dead_code)]` and no callers, either remove it or replace the annotation with a comment naming the planned caller.

crates/solver-delivery/src/implementations/evm/nonce.rs
nahimterrazas: If local cache says n, but real chain nonce is n + 10, we do not loop through 10 bad nonces. The submit path detects nonce too low, calls resync_nonce_for, fetches chain pending, resets to that value, retries once, and if it still fails, it returns DeliveryError::NonceTooLow

crates/solver-bridge/src/monitor.rs
nahimterrazas: Thanks! You are right, although this scenario is not common, I will create a follow-up PR for this.

crates/solver-delivery/src/implementations/evm/nonce.rs
nahimterrazas: because chain **_pending_** can lag, and the cache prevents immediate duplicate nonce reuse inside one solver process.

shahnami (approved):

PR #321, opened 2026-04-28
Add live fill estimate feature to gas configuration

nahimterrazas:
- Introduced a new boolean field  in the  struct to control the use of  for quote-time pricing.
- Updated the  to initialize this field to true by default.
- Enhanced the  to utilize the live fill estimate when enabled, allowing for more accurate gas cost calculations during quote processing.
- Adjusted related configurations and tests to ensure compatibility with the new feature.

# Summary

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Live post-fill gas estimation (per-chain opt-in) for more accurate quote-time costs.
  * Per-chain fee policy support with EIP-1559 tuning and configurable strategies.
  * Delivery layer exposes richer fee/estimation APIs for overrides and state-based gas checks.

* **Improvements**
  * Gas tracking extended to include post-fill and pre-claim phases for finer cost breakdowns.
  * Admin API and signed update flow expanded to configure live-estimate chains and new gas fields.

* **Documentation**
  * Added fee-policy and runtime guidance for enabling/validating live post-fill estimation.


crates/solver-core/src/engine/cost_profit.rs
```diff
@@ -37,6 +40,16 @@ pub enum CostProfitError {
...
 
+/// Left-pad a 20-byte EVM address into `bytes32`, right-aligned (occupies
+/// `bytes32[12..32]`). Caller passes raw 20 bytes; on a different length we
+/// truncate / zero-extend so we never panic.
+fn address20_to_bytes32(addr20: &[u8]) -> FixedBytes<32> {
+	let mut buf = [0u8; 32];
+	let len = addr20.len().min(20);
+	buf[32 - len..].copy_from_slice(&addr20[..len]);
+	FixedBytes::<32>::from(buf)
+}
```
shahnami: Seems like something that would already exists or can be done with an Alloy helper?

crates/solver-core/src/engine/mod.rs
```diff
@@ -105,6 +105,18 @@ pub struct SolverEngine {
...
 
+/// Returns true when the error string indicates a transient post-fill failure
+/// — one that may succeed on a later attempt without any code or order
+/// changes. Marking such orders as `Failed` would lock in a real loss
+/// because the Fill has already settled on chain; only the claim half
+/// remains, and it just needs another attempt with a healthier signer.
+///
+/// Currently identifies:
+/// - Insufficient native gas (balance changes outside the solver's view).
+fn is_transient_postfill_error(error_msg: &str) -> bool {
```
shahnami: The `contains("Insufficient native gas")` check is matching against a dynamically formatted string (`format!("Failed to handle PostFillReady: {e}")`), so if the underlying error's `Display` output ever changes, wording, capitalization, different wrapping, this silently returns `false`. An order where the fill has already settled on-chain would get marked `Failed`, locking in a real loss. Would it make sense to return a typed error from `handle_post_fill_ready` with an `InsufficientGas` variant and match on that instead?

crates/solver-core/src/engine/cost_profit.rs
```diff
@@ -1089,6 +1176,215 @@ impl CostProfitService {
...
+	///   the static config default — safe.
+	async fn build_fill_tx_for_quote(
+		&self,
+		request: &GetQuoteRequest,
+		_context: &ValidatedQuoteContext,
+		resolved_amounts: &std::collections::HashMap<InteropAddress, TokenAmountInfo>,
+		config: &Config,
+		solver_address: &Address,
+	) -> Result<Transaction, APIError> {
+		let output = request
```
shahnami, line 1697: This uses only the first output, so multi-output intents get an underestimated fill gas figure, the actual fill would cost more than the quote priced. Would it be worth either summing estimates across all outputs, or returning an error early when `outputs.len() > 1`?

crates/solver-core/src/engine/cost_profit.rs
```diff
@@ -37,6 +40,16 @@ pub enum CostProfitError {
 	Storage(#[from] solver_storage::StorageError),
 }
 
+/// Left-pad a 20-byte EVM address into `bytes32`, right-aligned (occupies
+/// `bytes32[12..32]`). Caller passes raw 20 bytes; on a different length we
+/// truncate / zero-extend so we never panic.
+fn address20_to_bytes32(addr20: &[u8]) -> FixedBytes<32> {
```
shahnami: Looks like this is exactly what Alloy's `Address::into_word()` does. Parsing the raw bytes to `alloy_primitives::Address::from_slice(bytes)` first gives you the right-aligned padding for free and would let you drop this function entirely.

crates/solver-config/src/lib.rs
```diff
@@ -322,12 +348,22 @@ fn default_rate_buffer_bps() -> u32 {
 	14 // 0.14%
 }
 
+fn default_live_fill_estimate_enabled() -> bool {
+	true
```
shahnami: nit: `default_true()` already exists in this file at line 144 and does the same thing — `#[serde(default = "default_true")]` on the new field would avoid the duplicate.

crates/solver-core/src/engine/cost_profit.rs
```diff
@@ -1089,6 +1176,215 @@ impl CostProfitService {
...
+			context: Vec::<u8>::new().into(),
+		};
+
+		let filler_data = address20_to_bytes32(&solver_address.0).to_vec();
+
+		// `fillDeadline` must be a near-future timestamp, but not too far ahead:
+		// production OutputSettlers reject both `0` (selector 0x9f3ddb90 =
+		// `FillDeadline()` — observed) AND values past their max-fill-window
+		// (which on OIF is typically the same order of magnitude as
+		// `api.quote.fill_deadline_seconds`, ~30 min). Use a tight `now + 60s`
```
shahnami: nit: The 60-second window is hardcoded; a settler that enforces a longer minimum fill-deadline will reject the `estimateGas` call and the live estimate silently falls back to static on every quote. Worth deriving from `config.api.quote.fill_deadline_seconds` so it stays in sync with the actual quote window.

shahnami (commented):
Nice work overall, spotted a few things worth addressing before merging.

crates/solver-core/src/engine/cost_profit.rs
nahimterrazas: We are not working now with output > 1

shahnami (approved):

PR #326, opened 2026-05-05
improve architecture testing

nahimterrazas:
# Summary

## Checklist


crates/solver-e2e-tests/src/lib.rs
```diff
@@ -546,6 +657,58 @@ pub fn amount_with_decimals_helper(whole: u64) -> U256 {
...
+
+pub fn assert_reverted_receipt(receipt: &TransactionReceipt, context: &str) {
+	assert!(
+		!receipt.status(),
+		"{context}: expected transaction to revert"
+	);
+}
+
+pub fn assert_balance_delta(before: U256, after: U256, expected: U256, context: &str) {
+	assert_eq!(after - before, expected, "{context}");
```
shahnami: This will panic with attempt-to-subtract-with-overflow whenever `after < before`, e.g. checking a sender's balance after a transfer. The function name doesn't promise direction, so the next test author who reaches for it can easily trip it. Could you switch to `after.checked_sub(before).expect("balance decreased")`, or take a signed delta? Worth noting nothing calls this yet, so this is mostly about not leaving a footgun in place.

crates/solver-e2e-tests/src/lib.rs
```diff
@@ -157,8 +157,44 @@ pub struct Harness {
 	_tempdir: TempDir,
 }
 
+#[derive(Debug, Clone)]
+pub struct HarnessOptions {
+	pub user_token_a_mint: U256,
+	pub solver_token_b_mint: U256,
+	pub use_false_oracle: bool,
+	pub use_reverting_output_settler: bool,
+	pub enable_permit2: bool,
```
shahnami: `enable_permit2` is declared here and seeded as `false` by `Default`, but I can't find any read of it in the harness, `spawn_solver`, or `build_seed_overrides`. A test author who sets it to `true` expecting Permit2 wiring will silently get the same flow as `false`. Could you either delete it, or if it's staged for a follow-up PR, add a `// TODO(phase 2)` style comment so the intent is explicit?

crates/solver-e2e-tests/src/lib.rs
```diff
@@ -157,8 +157,44 @@ pub struct Harness {
...
+	pub enable_admin_api: bool,
+	pub admin_redis_url: Option<String>,
+	pub enable_offchain_api: bool,
```
shahnami: Same situation as `enable_permit2` above: declared in `HarnessOptions` and `Default`, but never read. Either remove it or annotate it as a placeholder for a follow-up so the next person doesn't waste time chasing why their `enable_offchain_api: true` did nothing.

crates/solver-bridge/src/lib.rs
```diff
@@ -440,9 +440,53 @@ impl BridgeService {
...
+			)));
+		}
+		implementations.insert(name, factory);
+		Ok(())
+	}
+
+	pub fn all() -> Vec<(&'static str, BridgeFactory)> {
+		let implementations = Self::registry()
+			.lock()
+			.expect("bridge registry lock poisoned");
```
shahnami: `register` returns `BridgeError::Config` on a poisoned lock, but `all` here uses `.expect`. Since `get_all_implementations()` calls `all()` from the production solver bootstrap path, a panicked `register` permanently DOSes that path. Could you make `all()` return `Result<Vec<_>, BridgeError>` symmetric with `register`, or pick one consistent failure mode? Mixing the two means callers can't predict what they'll get.

crates/solver-e2e-tests/src/lib.rs
```diff
@@ -546,6 +657,58 @@ pub fn amount_with_decimals_helper(whole: u64) -> U256 {
...
+					|| debug.contains("insufficient")
+					|| debug.contains("allowance")
+					|| debug.contains("transfer amount exceeds"),
+				"{context}: expected revert/preflight failure, got {message}; debug: {debug}"
+			);
+		},
+	}
+}
+
+pub fn assert_reverted_receipt(receipt: &TransactionReceipt, context: &str) {
```
shahnami: `assert_reverted_receipt` is exported but I couldn't find a caller. Worth deleting until a test reaches for it; easier to add later than to maintain a public surface nothing exercises.

crates/solver-e2e-tests/src/lib.rs
```diff
@@ -713,13 +891,33 @@ async fn mint_erc20(
 // keyed by `chain_id` — NOT a HashMap. Hand-writing JSON in the runtime
 // `Config` shape (a map) is what was breaking the previous run.
 
+#[derive(Debug, Clone, Default)]
```
shahnami: `RebalanceOverridesBuilder` and the wrapper `build_seed_overrides_with_rebalance` are only ever invoked with `Default::default()` from the single caller in `build_seed_overrides`, so they don't actually buy anything yet. The Phase 2 comment up in `HarnessOptions` is fine motivation, but the indirection adds review surface without a test that needs it. Could you inline the `rebalance: rebalance.config` line back into `build_seed_overrides` and reintroduce the builder when Phase 2 actually customizes it?

crates/solver-e2e-tests/src/lib.rs
```diff
@@ -395,6 +458,51 @@ impl Harness {
 		}
 	}
 
+	/// Assert that no matching event appears within `timeout`.
+	pub async fn await_no_event<E: SolEvent>(
```
shahnami, line 449: Around 40 lines here mirror `await_event` above (block-window math, filter construction, polling loop). The risk with two parallel pollers is drift: someone fixes a block-window bug in one and forgets the other. Worth extracting a `poll_once` helper that returns `Option<(E, Log)>`, then both methods reduce to a small wrapper around it (`Err` on `Some` for `await_no_event`, `Err` on deadline for `await_event`).

crates/solver-e2e-tests/src/lib.rs
```diff
@@ -943,3 +1167,101 @@ pub fn nonce_from_seed(seed: &str) -> U256 {
 	U256::from_be_bytes::<32>(keccak256(seed.as_bytes()).into())
 }
 
+pub struct StandardOrderBuilder<'a> {
```
shahnami, line 1170: Small thing: adding `#[must_use = "call .build() to produce a StandardOrder"]` to the struct would catch the `StandardOrderBuilder::happy_path(&h, "x").amount_in(amt);` typo (forgot `.build()`) at compile time.

crates/solver-e2e-tests/src/lib.rs
```diff
@@ -395,6 +458,51 @@ impl Harness {
...
+			let filter = Filter::new()
+				.address(address)
+				.event_signature(E::SIGNATURE_HASH)
+				.topic1(order_id)
+				.from_block(from_block)
+				.to_block(BlockNumberOrTag::Number(head));
+
+			let logs = provider.get_logs(&filter).await.context("get_logs")?;
+			if let Some(log) = logs.into_iter().next() {
+				E::decode_log(&log.inner).context("decode unexpected event")?;
```
shahnami: Minor: the `E::decode_log(&log.inner)?;` here discards the decoded value, so the call exists only to assert the bytes parse. A `let _ =` plus a one-line comment would make the intent explicit, or you could drop the decode entirely since the topic match already proves the event is present.

scripts/e2e/solver_all_with_redis.sh
```diff
@@ -0,0 +1,39 @@
...
+redis_port="${E2E_REDIS_PORT:-6379}"
+redis_image="${E2E_REDIS_IMAGE:-redis:7-alpine}"
+
+if ! command -v docker >/dev/null 2>&1; then
+  echo "docker is required when REDIS_URL is not already set" >&2
+  exit 1
+fi
+
+if [ -n "${REDIS_URL:-}" ]; then
+  exec scripts/e2e/solver_all.sh
```
shahnami: Tiny portability thing: `scripts/e2e/solver_all.sh` here (and again at line 32) is relative to `$PWD`, so the wrapper only works from the repo root. `"$(dirname "$0")/solver_all.sh"` makes it directory-independent.

crates/solver-e2e-tests/tests/failure_e2e_onchain.rs
```diff
@@ -0,0 +1,101 @@
...
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+#[ignore = "requires Anvil + oif-contracts/out; opt-in via --ignored"]
+async fn expired_order_open_fails() -> anyhow::Result<()> {
+	let h = Harness::boot().await?;
+
+	let amount_in = amount_with_decimals_helper(1_000);
+	let order = StandardOrderBuilder::happy_path(&h, "e2e-expired-order")
+		.amount_in(amount_in)
+		.expires(unix_now_plus(0).saturating_sub(10))
```
shahnami: Tiny readability thing: `unix_now_plus(0).saturating_sub(10)` is computing 'now minus 10s', and the saturating guard handles a case that can't happen (Unix epoch never wraps in `u32` going down by 10). A `unix_now_minus(10)` helper or just `unix_now_plus(0) - 10` reads cleaner.

crates/solver-e2e-tests/src/lib.rs
```diff
@@ -889,11 +1104,16 @@ fn spawn_solver(
 		// CoinGecko for our fake TOKA/TOKB. The mock impl has TOKA/USD and
 		// TOKB/USD prices baked in — see `MockPricing::new`.
 		.env("PRICING_PRIMARY", "mock")
+		.env("JWT_SECRET", "solver-e2e-admin-jwt-secret")
```
shahnami: Minor: `JWT_SECRET` is set every spawn even when `enable_admin_api: false`. Gating it inside an `if options.enable_admin_api { ... }` block alongside the `REDIS_URL` env at lines 1112-1114 keeps non-admin runs from carrying admin-only config.

shahnami (changes requested):
Solid expansion of the e2e scenarios. There's one correctness bug in a helper, a few unused public surfaces that could mislead future test authors, plus some smaller polish items..

luiz-lvj (approved):
LGTM

PR #331, opened 2026-05-06
E2e test ci

nahimterrazas:
# Summary

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a health endpoint that reports startup readiness (including approvals and blocked signers).
  * Admin API now supports role-based whitelist management with endpoints to list and set roles; finer-grained admin read vs write scopes.

* **Chores**
  * Added an end-to-end CI workflow and improved CI triggers, caching, concurrency, and rollup behavior.
  * Updated fee estimation guidance to use a larger base-fee buffer (reflected in docs).


.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,162 @@
...
+      - docs/**
+      - "**.md"
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - docs/**
+      - "**.md"
+  workflow_dispatch:
+
```
tirumerla, line 22: Let's add concurrency to make sure there are no queues for the build

```
concurrency:
    group: ${{ github.workflow }}-${{ github.ref }}
    cancel-in-progress: true
```

.github/workflows/ci.yaml
tirumerla: why not remove it here as well?

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,162 @@
...
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          repository: openintentsframework/oif-contracts
+          path: oif-contracts
+          # forge-std, permit2, the-compact, openzeppelin-contracts, broadcaster
+          submodules: recursive
+          # Tracks upstream main. Replace with a commit SHA if you need
+          # strict reproducibility across CI runs (cache key is keyed on the
+          # resolved SHA either way, so upstream changes invalidate cleanly).
+          ref: main
```
tirumerla, line 56: This is fine for now, probably better to have determinism by pinning to commit sha instead of main

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,162 @@
...
+        with:
+          version: stable
+
+      - name: Build oif-contracts
+        if: steps.forge-cache.outputs.cache-hit != 'true'
+        run: forge build
+        working-directory: oif-contracts
+
+      - name: Prepare Rust toolchain
+        uses: ./.github/actions/prepare
```
tirumerla: may be adding a shared hash key would be better so that all shards can retrieve compiled wihtout rebuilding it again

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,162 @@
...
+      # `fail-on-cache-miss` so a missing upstream save fails loudly rather
+      # than silently triggering a rebuild path that doesn't exist here.
+      - name: Restore forge build output
+        uses: actions/cache/restore@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
+        with:
+          path: |
+            oif-contracts/out
+            oif-contracts/cache
+          key: forge-build-${{ runner.os }}-${{ needs.build.outputs.forge-cache-key }}
+          fail-on-cache-miss: true
```
tirumerla: we can remove this and use another step to checkout if the cache hit doesn't happen, that way you will fallback to rebuilding instead of failing all the shards

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,162 @@
...
+
+      # anvil is a runtime dependency; forge isn't needed since out/ is restored.
+      - name: Install Foundry (anvil)
+        uses: foundry-rs/foundry-toolchain@de808b1eea699e761c404bda44ba8f21aba30b2c # v1.4.0
+        with:
+          version: stable
+
+      - name: Prepare Rust toolchain
+        uses: ./.github/actions/prepare
```
tirumerla: same as https://github.com/openintentsframework/oif-solver/pull/331/changes#r3197944259

tirumerla (commented):
Added some nits and improvements

.github/workflows/e2e.yaml
luiz-lvj: yeah, this makes sense, added

.github/workflows/ci.yaml
luiz-lvj: removed.

.github/workflows/e2e.yaml
luiz-lvj: Yep, this makes sense. Just set shared-key: e2e-suite on both. The first run after this still has a cold cache, but subsequent runs should hit.

.github/workflows/e2e.yaml
luiz-lvj: Switched it

tirumerla (approved):

PR #337, opened 2026-05-11
Eth bridge native rebalance

nahimterrazas:
## Summary by CodeRabbit

* **New Features**
  * Per-pair bridge routing with route-aware transfers and native wrap/unwrap orchestration; native flows now include wrap_pending / unwrap_pending states.
  * ABI/event support added for native wrapper and vault flows to enable reliable unwrap/redeem detection.

* **Improvements**
  * Stronger crash-window protections, automatic resubmission/escalation, and startup preflight validation for route-configured pairs.

* **Documentation**
  * Expanded rebalance docs: native-asset guidance, route schema, lifecycle/status updates.

* **Tests**
  * Added routing and native-flow test coverage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openintentsframework/oif-solver/pull/337)


pepebndc (commented):
Two concerns about `RebalanceMonitor::logical_balance` in `crates/solver-bridge/src/monitor.rs`:

**1. Parse errors silently fall back to zero balance**

```rust
let native_u = U256::from_str_radix(&native, 10).unwrap_or(U256::ZERO);
let wrapper_u = U256::from_str_radix(&wrapper_balance, 10).unwrap_or(U256::ZERO);
```

If `delivery.get_balance` ever returns a non-decimal representation (hex from a provider quirk, RPC change), the balance silently reads as `0`. The threshold check would then fire a "balance below target — rebalance!" trigger on a phantom-empty wallet, and the bridge would only fail later with `InsufficientNativeGas` after wasting a wrap/approve cycle. Suggest propagating the parse error as `DeliveryError` (or at minimum `tracing::warn!`) so a malformed balance is observable instead of silently equivalent to an empty wallet.

**2. Logical balance inflates apparent native by wrapper holdings; auto-trigger can over-bridge**

`logical_balance` returns `native + wrapper.balanceOf` for native sides, but the auto-trigger downstream computes `amount_to_bridge = balance - target` and the wrap step requires `value: amount` in real native ETH. Scenario:

- A wrap tx got dropped from the mempool earlier and `WRAP_MISSING_CHECKS_MAX` cleared the hash → WETH balance still high, native lower.
- Auto-trigger reads logical balance as high, computes a surplus, schedules `WETH.deposit{value: amount}`.
- `deposit` reverts / hits `InsufficientNativeGas` because we don't actually hold that much native.

Not catastrophic — the `InsufficientNativeGas` arm escalates cleanly — but produces noisy false-positive transfer attempts and `NeedsIntervention` entries. Suggest either subtracting in-flight `WrapPending` amounts before computing the trigger, or capping the bridge amount by `min(logical_balance - target, native_balance - min_native_gas_reserve)`.

pepebndc (approved):
lgtm

PR #355, opened 2026-05-18
fix(security): filter OutputFilled logs by emitter and decode Mandate…

nahimterrazas:
…Output

  extract_fill_details_from_logs in both broadcaster and hyperlane previously
  matched only on event topic and indexed orderId, so any contract executed
  during the fill transaction could emit a forged OutputFilled log with an
  attacker-chosen solver and timestamp and steer the solver's post-fill
  submission and stored payload hash toward the spoofed payload.

  Both helpers now:
    - resolve the expected outputs[i] from the order by destination chain,
    - require log.address == the order-bound output settler (left-padded EVM address derived from outputs[i].settler bytes32, with a check that the upper 12 bytes are zero), type and compare MandateOutput field-by-field against the order.

# Summary

## Checklist

## Summary by CodeRabbit

# Release Notes

* **Bug Fixes**
  * Strengthened order fulfillment verification by implementing comprehensive validation of fulfillment event logs against order details, ensuring critical fields (recipient, token, amount, and chain data) match expectations and rejecting mismatches or forged logs.

* **Tests**
  * Expanded test coverage for fulfillment log validation, including scenarios for rejected forged events and mismatched data.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openintentsframework/oif-solver/pull/355?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/solver-order/src/implementations/standards/_7683.rs
+++ b/crates/solver-order/src/implementations/standards/_7683.rs
@@ -419,8 +419,11 @@ impl OrderInterface for Eip7683OrderImpl {
 		let user_address = hex_to_alloy_address(&order_data.user)
 			.map_err(|e| OrderError::ValidationFailed(format!("Invalid user address: {e}")))?;
 
-		// Parse oracle address
-		let oracle_address = hex_to_alloy_address(&fill_proof.oracle_address)
+		// Security: use the order-bound input oracle from canonical order_data, not
+		// the FillProof's oracle_address. The settlement service's get_attestation
+		// guard enforces that these match, so this preserves the order identity at
+		// claim time even if a future settlement impl regresses.
+		let oracle_address = hex_to_alloy_address(&order_data.input_oracle)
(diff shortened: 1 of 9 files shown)


pepebndc (changes requested):
Re-reviewed after merging fix PR #360. Production code is correct in both files — one test gap to close before merge:

**Missing test in `hyperlane.rs`**

`test_extract_fill_details_rejects_diverged_final_amount` was added to `broadcaster.rs` but not to `hyperlane.rs`. The divergence guard is correctly implemented in both files, but the test coverage is asymmetric. Given this is a security-critical path, both settlement implementations should have matching test coverage.

The fix is small — copy the test from `broadcaster.rs` into the `#[cfg(test)]` block in `hyperlane.rs`, adjusting any helper calls to match the hyperlane test setup. Everything else (Fix 1 final_amount divergence check, Fix 2 ABI decode warn) looks good in both files.

pepebndc (approved):

PR #356, opened 2026-05-18
Enhance oracle binding security in settlement implementations

nahimterrazas:
- Updated Eip7683OrderImpl to use the order-bound input oracle from canonical order data instead of FillProof's oracle address, ensuring order identity is preserved during claims.
- Introduced shared helpers for parsing order-bound oracles in the new oracle_binding module.
- Implemented validation methods for bound input and output oracles in DirectSettlement and HyperlaneSettlement, enforcing security checks against configured oracles.
- Added tests to verify correct behavior of oracle validation and rejection of mismatched oracles in settlement services.

# Summary

## Checklist

### DIFF
--- a/crates/solver-order/src/implementations/standards/_7683.rs
+++ b/crates/solver-order/src/implementations/standards/_7683.rs
@@ -419,8 +419,11 @@ impl OrderInterface for Eip7683OrderImpl {
 		let user_address = hex_to_alloy_address(&order_data.user)
 			.map_err(|e| OrderError::ValidationFailed(format!("Invalid user address: {e}")))?;
 
-		// Parse oracle address
-		let oracle_address = hex_to_alloy_address(&fill_proof.oracle_address)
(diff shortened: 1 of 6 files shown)


pepebndc (approved):
Re-reviewed after merging fix PR #359. Both concerns addressed:

- `readiness_internal` now calls `validate_bound_output_oracle(order, destination_chain)` and compares against `submission.remote_oracle`, returning `PermanentFailure` on mismatch. Test `test_readiness_internal_rejects_remote_oracle_mismatch` covers the exact scenario (in-config but not signed oracle).
- `parse_bound_input_oracle` rejects zero-address input oracle symmetrically with the output-oracle side. Test `test_parse_bound_input_oracle_rejects_zero_oracle` confirms.

The previously-flagged `generate_pre_claim_transaction` config-only check appears to be covered by the broader oracle-binding refactor — no residual config-only path found.

Approving. 🟢

PR #364, opened 2026-05-28
pc-04 - compact-signature-canonicalization

nahimterrazas:
# Summary

## Checklist

## Release Notes

* **Bug Fixes**
  * Enhanced ResourceLock signature validation to strictly verify off-chain signatures on intent submission; requests with invalid or malformed signatures now return `400 Bad Request` with clear rejection messages.
  * Improved signature extraction error handling to surface validation failures as API errors instead of silently assuming success.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openintentsframework/oif-solver/pull/364?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/solver-discovery/src/implementations/offchain/_7683.rs
+++ b/crates/solver-discovery/src/implementations/offchain/_7683.rs
@@ -61,6 +61,7 @@ use solver_types::{
 	api::PostOrderRequest,
 	bytes32_to_address, create_http_provider, current_timestamp, normalize_bytes32_address,
 	standards::eip7683::{
+		compact_signatures::decode_compact_signatures,
 		interfaces::{IInputSettlerCompact, IInputSettlerEscrow, SolMandateOutput, StandardOrder},
 		GasLimitOverrides, LockType, MandateOutput,
 	},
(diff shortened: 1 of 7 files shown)


pepebndc (approved):
Validated against scan finding **C-04** — fully fixed.

New `decode_compact_signatures` (`crates/solver-types/src/standards/eip7683/compact_signatures.rs`) replaces the fixed-offset sponsor-signature extraction with a strict canonical ABI decoder: enforces sponsor offset == 64, contiguous allocator offset, in-bounds length/data/padding (`checked_add`, usize-range word guard), and no trailing data. `extract_signature` now returns `Result` and the validator path propagates with `?`, so a non-canonical payload is rejected (BadRequest) instead of being silently sliced at a fixed 0x60 offset. Because the decoder is strictly stronger than Solidity `abi.decode`, any payload it accepts decodes to the same sponsor bytes on-chain — closing the off-chain/on-chain extraction mismatch. Wired into the live ResourceLock path (`handle_order` → `validate_signature` → `extract_signature`), with a defense-in-depth gate in discovery `handle_intent_submission`. Tests exercise the actual shifted-offset exploit and assert rejection.

Note: this commit (`83790f9c`) is identical to #369's first commit, so the two share the C-04 fix — only one needs to merge. M-96 (non-canonical v/s in `recover_signer`) is a separate finding, not in this PR's scope. LGTM for C-04.

PR #372, opened 2026-06-01
h-24 feat: add exact output swap amounts calculation and corresponding test

nahimterrazas:
# Summary

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed exact-output swap results to include the specified output amount and correct token decimals while still showing the calculated input, ensuring displayed swap details and confirmations are accurate.

* **Tests**
  * Added a test to verify exact-output swaps include the known output amount in results.

### DIFF
--- a/crates/solver-core/src/engine/cost_profit.rs
+++ b/crates/solver-core/src/engine/cost_profit.rs
@@ -487,6 +487,14 @@ impl CostProfitService {
 						.map_err(|e| CostProfitError::Calculation(e.to_string()))?;
 
 						output_usd_values.push(usd_value);
+						calculated_amounts.insert(
+							output.asset.clone(),
+							TokenAmountInfo {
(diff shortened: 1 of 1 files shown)


pepebndc (approved):
Validated against scan finding **H-24** — fully fixed.

`calculate_swap_amounts` now inserts the requested output asset (keyed by `output.asset`) into the ExactOutput result map. Both synthetic builders (`build_fill_tx_for_quote`, `build_post_fill_tx_for_quote`) read exactly that key and previously failed deterministically — silently falling back to static gas units and underpricing quotes. With the entry present, live estimateGas runs. Added RED-GREEN test confirms the output entry is present (would be `None` without the fix).

Minor follow-up: builder errors are still log-and-continue when live estimation is enabled (the finding's optional 'treat as hard error' suggestion). Independently verified. LGTM.

PR #373, opened 2026-06-01
h-26 Implement validation to reject multiple cross-chain outputs

nahimterrazas:


pepebndc (approved):
Validated against scan finding **H-26** — fully fixed.

`validate_order` now rejects any order with more than one distinct cross-chain output (`seen_chains.len() > 1`), on the sole intake chokepoint (`validate_and_create_order`). Combined with existing same-chain and duplicate-chain rejection, exactly one cross-chain output remains — which is what `generate_fill_transaction`'s single `.find()` actually fills, so the partial-fill/unclaimable mismatch is unreachable. This is the finding's explicitly-endorsed remediation (reject multi-output until full multi-output execution exists). Independently verified. LGTM.

PR #374, opened 2026-06-01
h-23 fix: validate resource lock signatures before capacity checks

nahimterrazas:
# Summary

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Signature validation now runs earlier in request processing so invalid signatures are rejected before any capacity/delivery checks, preventing unnecessary downstream operations.

* **Tests**
  * Added tests and helpers to assert early signature validation and confirm capacity/delivery paths are not invoked for invalid requests.

### DIFF
--- a/crates/solver-service/src/server.rs
+++ b/crates/solver-service/src/server.rs
@@ -784,6 +784,19 @@ async fn validate_intent_request(
 			details: None,
 		})?;
 
+	// EIP-712 signature validation for ResourceLock orders
+	let requires_validation = state
+		.signature_validation
+		.requires_signature_validation(standard, &lock_type);
+
+	if requires_validation {
+		let config = state.config.read().await;
+		state
+			.signature_validation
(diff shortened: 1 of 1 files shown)


pepebndc (approved):
Validated against scan finding **H-23** — fully fixed.

EIP-712 signature validation is relocated in `validate_intent_request` (server.rs) to run *before* `ensure_user_capacity_for_order`. An invalid ResourceLock signature is now rejected after a single constant-cost DOMAIN_SEPARATOR RPC + local ECDSA recovery, before the attacker-controlled per-input capacity RPC loop. On-by-default for all ResourceLock orders. Added test asserts the capacity-check RPC is never reached on bad signatures.

Minor follow-up: the finding's secondary suggestion (input-length / request-body caps) is not implemented — defense-in-depth, doesn't reopen the amplification vector. Independently verified. LGTM.

PR #375, opened 2026-06-01
h-22 fix: reject unsupported intent assets before fill simulation

nahimterrazas:
# Summary

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added pre-simulation validation for orders that checks token metadata resolution, callback policies, and configuration constraints before running fill simulations to improve efficiency.

* **Tests**
  * Added tests to verify invalid orders are properly rejected during validation and skipped before expensive simulation operations.

### DIFF
--- a/crates/solver-core/src/engine/cost_profit.rs
+++ b/crates/solver-core/src/engine/cost_profit.rs
@@ -1237,6 +1237,18 @@ impl CostProfitService {
 		Ok(())
 	}
 
+	/// Returns true if `recipient_interop_hex` is permitted by the configured
+	/// callback whitelist. An empty whitelist means "allow all callbacks",
+	/// matching the quote-time contract in solver-service's
+	/// `validate_callback_whitelist`.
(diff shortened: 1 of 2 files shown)


pepebndc (approved):
Validated against scan finding **H-22** — fully fixed.

New `CostProfitService::validate_before_fill_simulation` runs in `IntentHandler::handle` immediately before `generate_fill_transaction`/`simulate_callback_and_estimate_gas`, performing only cheap local checks (token allowlist via in-memory `get_token_info`, callback calldata hex/size cap, callback-whitelist policy) and emitting `OrderEvent::Skipped` on failure — so unsupported assets never reach `eth_estimateGas`. Tests assert `generate_fill_transaction.times(0)`.

Minor follow-up (out of scope, default-off): the quote-path `try_live_fill_estimate` lacks an equivalent pre-check and fails open on token lookup, but is gated behind `live_fill_estimate_enabled=false`. Independently verified. LGTM.

PR #377, opened 2026-06-01
h-4-5,13 fix: harden public api intake and startup nonce probing

nahimterrazas:
## Summary by CodeRabbit

* **New Features**
  * Added configurable API rate limiting with per-IP request throttling.
  * Added request payload size limits on API endpoints.
  * Added quote endpoint concurrency limiting to prevent overload.
  * Implemented per-chain concurrency limits for live gas estimation.

* **Bug Fixes**
  * Improved error handling for saturated intent queues (returns 503 Service Unavailable instead of potential hangs).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -75,6 +75,7 @@ jobs:
       - msrv
       - rustfmt
       - clippy
+      - test
     runs-on: ubuntu-latest
     steps:
       - name: Harden the runner (Audit all outbound calls)
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -171,6 +171,7 @@ jobs:
     needs:
       - changed_files
     runs-on: ubuntu-latest
+    timeout-minutes: 25
     services:
       redis:
         image: redis:7-alpine
@@ -214,8 +215,6 @@ jobs:
       # Unit tests coverage
       - name: Run Unit Tests and Generate Coverage Report
         env:
-          LLVM_PROFILE_FILE: unit-%p-%m.profraw
-          RUSTFLAGS: -Cinstrument-coverage
           RUST_TEST_THREADS: 1
         run: cargo llvm-cov --locked --workspace --exclude solver-e2e-tests --lcov --output-path unit-lcov.info
 
--- a/README.md
+++ b/README.md
@@ -458,6 +458,7 @@ Token arrays may be empty at boot and can be populated later via admin APIs.
   "gas_buffer_bps": 1500,
   "commission_bps": 20,
   "rate_buffer_bps": 14,
+  "resource_lock_enabled": false,
   "networks": [
     {
       "chain_id": 11155420,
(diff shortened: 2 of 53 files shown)


pepebndc (changes requested):
Validated against scan findings **H-04, H-05, H-13**. H-04 is robustly fixed, but **H-05 and H-13 still ship exploitable by default** — requesting changes.

**H-04 (unbounded intent queue) — ✅ fixed.** Bounded `mpsc::channel(1024)` replaces the unbounded channel; the off-chain `/intent` handler uses `try_send` and returns HTTP 503 on a full queue; on-chain path uses `send().await` for backpressure. Always-on, fail-closed. 

**H-05 (default-on live-fill `eth_estimateGas` amplification) — ❌ blocking.** The root cause is untouched: `live_fill_estimate_enabled` still defaults to `true` (`operator_config.rs:457`, `seeds/types.rs:234`), and no estimateGas concurrency cap / caching / time-variant-field normalization was added. The only relevant new protection (per-IP rate limiter) is **default-off**. A stock solver still performs unauthenticated, unthrottled, uncached destination-chain `eth_estimateGas` per `/quotes` request.

**H-13 (unauthenticated `/quotes` RPC+storage DoS) — ⚠️ major.** Good primitives added (`DefaultBodyLimit`, optional JWT scope, optional rate limiter, `/health`), but the two primary protections are **default-off**: quote auth is gated on `orders_auth_enabled` (defaults `false`, `config_merge.rs:508`) and `rate_limiting` defaults to `None` (`config_merge.rs:2229`). No per-client / global cap on persisted `StoredQuote` was added, so the storage-amplification sub-case is unmitigated even with rate limiting on.

**Asks:**
1. H-05: default the live-fill estimate **off** for the public/unauthenticated quote path (or require auth), and/or add a per-chain estimateGas concurrency cap + short-TTL cache keyed on `(chain, to, calldata-hash)`.
2. H-13: ship safe defaults — enable rate limiting by default (or fail-closed if unset on a public endpoint), and add a per-client/global cap on persisted quotes.

**Minor (non-blocking):**
- `ApiRateLimiter.clients` map has no eviction of empty entries and no global bound (memory growth under varied source IPs); lock uses `.expect` (panics on poison); no `X-Forwarded-For` handling, so behind a proxy all clients collapse to one bucket.

pepebndc (changes requested):
**Re-review of the updated revision** (`e50659b` + merges). Big improvement — the prior **blocking** gap is resolved. Two **major** items remain before approve.

**Resolved since last review:**
- ✅ **H-04** still fully fixed (bounded `mpsc::channel(1024)` + `try_send`→503; on-chain path now `send().await` backpressure).
- ✅ Rate limiting is now **default-ON** (`config_merge.rs` → `rate_limiting: Some(RateLimitConfig::default())`, 600/min/IP), and a **per-chain live-estimate concurrency cap** (default 8, `LiveEstimateController`) is wired into the real `try_live_fill_estimate` path (fails soft to static gas). Quote route also gets a concurrency cap (32) + 1 MB body limit. This clears the prior "only protection was default-off" blocker. Rate-limiter eviction was added too.

**Remaining gaps (major):**
1. **H-05 root cause** — `live_fill_estimate_enabled` still defaults **TRUE** (`operator_config.rs:457`, `seeds/types.rs:234`, `builders/config.rs`). The finding's *primary* recommendation was to default this **off** for the public/unauthenticated quote path. As-is, a stock solver still performs unauthenticated destination `eth_estimateGas` per `/quotes` — now bounded (cap 8 + rate limit), but not eliminated. No caching/normalization of the time-variant `fillDeadline` was added (explicitly deferred), so identical logical requests stay non-coalescable. → default it off for public APIs (or require auth), and/or add a short-TTL cache keyed on `(chain, to, calldata-hash)`.
2. **H-13 storage amplification** — no per-client/global cap on persisted `StoredQuote`; `apis/quote/mod.rs::store_quotes` still calls `store_with_ttl` unconditionally per accepted quote (only indirectly bounded by rate limit). The finding explicitly asks to bound stored quotes per client. (Treating `/quotes` as public-but-protected is fine per the finding — the missing piece is the storage cap.)

**Minor:**
- Rate limiter keys solely on `ConnectInfo` peer IP — no `X-Forwarded-For`/`Forwarded` support, so behind a reverse proxy all clients collapse into one bucket (shared-budget DoS); falls back to `0.0.0.0` when absent. No max-cardinality bound on the per-IP map; `.expect` on lock poison in `ApiRateLimiter`/`LiveEstimateController`.

Once (1) and (2) are addressed I'm happy to approve. Nice progress.

pepebndc (approved):

PR #378, opened 2026-06-02
h-18-27-30: clamp legacy gas price and improve nonce management

nahimterrazas:
# Summary

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Legacy gas-price fallbacks are now capped by the active fee policy to produce consistent, bounded fee suggestions.
  * Nonce cache handling improved: it avoids moving below locally issued nonces and can reclaim a single definitively rejected nonce when safe, reducing stuck or conflicting submissions.
  * Fee-error classification expanded to better detect “max fee below block base fee” rejections.

### DIFF
--- a/crates/solver-delivery/src/implementations/evm/alloy.rs
+++ b/crates/solver-delivery/src/implementations/evm/alloy.rs
@@ -243,6 +243,14 @@ enum NonceCacheAction {
 	AttemptRollback { rejected_nonce: Option<u64> },
 }
 
+fn nonce_too_low_retry_nonce_cache_action() -> NonceCacheAction {
+	// A nonce-too-low result proves this nonce is already consumed or held.
+	// Even if a follow-up pending read is stale, do not reclaim it.
+	NonceCacheAction::AttemptRollback {
+		rejected_nonce: None,
+	}
+}
+
 /// Outcome of a raw-send attempt, after pre-sign persisted the hash.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(crate) enum RawSendVerdict {
@@ -2663,9 +2671,7 @@ impl DeliveryInterface for AlloyDelivery {
 						let cache_after = apply_nonce_cache_action(
 							mgr,
 							from,
-							NonceCacheAction::AttemptRollback {
-								rejected_nonce: tx_attempt.nonce,
-							},
+							nonce_too_low_retry_nonce_cache_action(),
 							pending_opt,
 						);
 						if let Some(pending) = pending_opt {
@@ -2677,7 +2683,7 @@ impl DeliveryInterface for AlloyDelivery {
 								cache_before = ?cache_before,
 								cache_after = ?cache_after,
 								error = %second_reason,
-								"nonce-too-low retry also returned nonce too low; nonce cache reconciled after scoped rollback"
+								"nonce-too-low retry also returned nonce too low; nonce cache kept forward-only"
 							);
 						} else {
 							tracing::warn!(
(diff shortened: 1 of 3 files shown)


pepebndc (changes requested):
Validated against scan findings **H-18, H-27, H-30**. H-30 and H-18 are correctly fixed, but **H-27 is only partially resolved** — requesting changes.

**H-30 (legacy gas cap) — ✅ fixed.** `clamp_legacy_gas_price_to_cap` is wired into both legacy fallback branches in `get_fee_params` (alloy.rs ~2721 and ~2753).

**H-18 (nonce-reuse safety) — ✅ fixed.** `reset_next_nonce` now returns `local.max(next)`, so a stale-pending rollback can't move the cache below the local high-water mark.

**H-27 (fee-cap/base-fee misclassification → nonce wedge) — ⚠️ major gap.** The reclassification to `DefinitelyRejected` is correct, but the rollback it now triggers is **neutralized by this same PR's H-18 clamp**. For the common single-in-flight-tx case (cache at N+1, chain `pending`=N because the tx never entered the pool), `reset_next_nonce` returns `max(N+1, N) = N+1`, so nonce N is never reclaimed. `next_nonce_for` uses the forward-only `set_next_nonce` and the drift monitor only logs — nothing frees the leaked nonce, so the **permanent nonce gap / liveness wedge the finding describes still occurs**. The PR's own renamed test `apply_nonce_cache_action_rollback_with_pending_preserves_high_water_cache` demonstrates the rollback is now a no-op (cache stays 101 when pending=100). The added H-27 tests only assert classification strings, not end-to-end nonce reclamation.

**Asks:**
1. Reconcile H-18 vs H-27: after a `DefinitelyRejected` (never-broadcast) outcome, the leaked nonce must actually be reclaimed for the lone-in-flight case. Consider scoping the rollback to the rejected attempt's nonce rather than relying on the high-water clamp, and add a test asserting the next `next_nonce_for` returns N (not N+1).

**Minor (non-blocking):**
- H-18's clamp removes the backward-reset that the nonce-too-low resync path still depends on; the comments at `alloy.rs:1851-1854` / `723` and the resync trace log now contradict actual behavior — please update or rescope.
- H-27 matches 3 substring phrasings rather than structured RPC error codes; other providers' fee-cap strings fall back to `Ambiguous`.

pepebndc (changes requested):
**Re-review of the updated revision.** The only change since my last review is a merge of `main` (no new fix commit), so the **H-27 gap still stands**, and a second look confirms a **regression** in the H-18 fix. Requesting changes.

- ✅ **H-30** fully fixed — `clamp_legacy_gas_price_to_cap` wired into both legacy fallback branches of `get_fee_params`; default `gas_price_cap=None` passes through, matching the EIP-1559 path. No issue.

- ⚠️ **H-27 still not remediated end-to-end (major).** Reclassifying the fee-cap/base-fee errors to `DefinitelyRejected` is correct and triggers a rollback — but that rollback is **neutralized by this PR's own H-18 clamp**. `apply_nonce_cache_action → reset_next_nonce(pending)` returns `max(local, pending)`; for the exact H-27 scenario (lone in-flight tx: cache `N+1`, chain `pending = N` because the tx never entered the pool), it returns `N+1`, so leaked nonce `N` is never reclaimed and the permanent nonce-gap / liveness wedge persists. The only backward-capable primitive (`reconcile_with_chain_pending`) has no production caller. The added H-27 tests assert only classification strings — the end-to-end nonce-reclamation path is untested.

- ⚠️ **Regression from the H-18 fix (major).** Making `reset_next_nonce` forward-only (`local.max(next)`) breaks the **nonce-too-low backward-resync** recovery path (`alloy.rs` ~1865) that explicitly depends on moving the cache *backward* to recover from ghost broadcasts. The comments there ("move the cache backward — the whole point of the resync"), the `cache_after = fresh_pending` trace, and the `next_nonce_for` doc are now false; `take_next` hands out the stale local value when `fresh_pending < local`.

**Asks:**
1. Distinguish "roll back a just-allocated, never-broadcast/rejected nonce" from "preserve in-flight nonces" — e.g. a dedicated rollback that targets the specific rejected nonce (or decrement-on-rejection), rather than reusing the forward-only `reset_next_nonce`. Add a test asserting a definitively-rejected lone tx reclaims `N` (next `next_nonce_for` returns `N`, not `N+1`).
2. Route the nonce-too-low resync through a backward-capable path (e.g. `reconcile_with_chain_pending`) and fix the now-false comments/traces.

**Minor:** H-27 still matches 3 English substrings rather than structured RPC error codes — brittle to other providers' wording.

pepebndc (approved):

PR #388, opened 2026-06-10
fix(C-06): atomic compact-deposit reservation accounting at engine acceptance

nahimterrazas:
# Summary

## Checklist

## Release Notes

* **New Features**
  * Added deposit reservation system for TheCompact resource-lock orders to ensure capacity availability
  * Added balance verification for Compact deposits during order intake

* **Improvements**
  * Enhanced safety validation of resource-lock order parameters
  * Improved order lifecycle management with automatic resource cleanup on terminal states
  * Better handling of concurrent order processing for deposit locks


pepebndc (approved):
Reviewed the reservation accounting paths. The core invariants hold: every intake path that takes a reservation releases it on `Skip`/store-failure, all early exits occur before the reservation is taken, the CAS + per-lock-id mutex prevents two concurrent orders from drawing on the same deposit, and holding reservations on `Failed(_)` until expiry (releasing only on `Finalized`) is the correct conservative choice given the on-chain idempotency invariant — held is always safe, early release on an ambiguous `Failed(Fill)` is not. Shared-store wiring between `IntentHandler` and `OrderStateMachine` is correct (`compact_reservations()` clones the Arc, so the lock-guards map is shared).

Non-blocking notes:
- `release_order` fail-fasts on the first per-lock error despite the "best-effort" comment at the call site; a multi-lock order can end up partially released (safe but inconsistent). Consider iterating all locks and collecting errors.
- `lock_guards` grows unboundedly (one entry per unique lock key, never shrunk). Bounded in practice, but a future cleanup candidate.
- The `terminal_transition_releases_compact_reservation` test constructs a second `CompactReservationStore` and only works because both share the same `Arc<StorageService>`; a comment would prevent confusion (the `intent.rs` test covers the real wiring).
- `reserve_one` allows a same-`order_id` re-reservation with a larger amount to pass against `reserved_by_others` only — fine while order data is immutable, worth documenting as an assumption.

PR #389, opened 2026-06-10
C04 intent in process

nahimterrazas:
# Summary

## Checklist


pepebndc (approved):
Reviewed with focus on the trust boundary, since this removes the discovery-side allocator authorization. Verified at PR head that the only production path to `DiscoveryService::submit_order` is `handle_order → validate_intent_request → submit_order_to_discovery` (server.rs), with the load-bearing comment documenting that `submit_order` is a trusted enqueue primitive. The removed allocator-auth code in `_7683.rs` was redundant duplication of `compact_allocator.rs`, which remains the single enforcement point via `validate_signature`. The `calls == 0` assertions in the gating tests prove rejected orders never reach discovery. Dropping the internal HTTP server (and reqwest/wiremock/axum deps from solver-discovery) is a clean simplification.

Non-blocking notes:
- `submit_order` runs `order_to_intent` (including an RPC call for the order ID) before checking `intent_sender.is_none()`; moving the check earlier would short-circuit the misconfigured case cheaper.
- `docs/offchain-discovery-security.md` is deleted; a short note in operator docs that port 8081 no longer exists/needs firewalling would help upgraders.

PR #390, opened 2026-06-10
h04-hyperlane-recovery-callback

nahimterrazas:
# Summary

## Checklist

## Summary by CodeRabbit

* **Improvements**
  * More consistent handling of transaction confirmations for PostFill and PreClaim, ensuring settlement callbacks are invoked and failures are surfaced.
  * Automated recovery and rebuilding of missing post-fill message tracking so attestation data can be recovered on demand.
  * Centralized receipt parsing for PostFill tracking.

* **Tests**
  * Added async tests covering recovery, tracker rebuild, and post-fill callback behavior.

### DIFF
--- a/crates/solver-core/src/handlers/transaction.rs
+++ b/crates/solver-core/src/handlers/transaction.rs
@@ -202,15 +202,13 @@ impl TransactionHandler {
 		// after persisting the chain-confirmed stage. A transient callback failure
 		// must not erase the fact that the transaction succeeded on-chain.
 		if let Some(order) = settlement_callback_order {
-			if let Ok(settlement) = self.settlement.find_settlement_for_order(&order) {
-				settlement
-					.handle_transaction_confirmed(&order, tx_type, &receipt)
-					.await
-					.map_err(|source| TransactionError::SettlementCallback {
-						stage: tx_type,
-						source,
-					})?;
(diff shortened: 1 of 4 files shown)


pepebndc (approved):
Reviewed the recovery paths. Both fixes check out: `recover_post_fill_state` correctly re-derives the message ID from the post-fill + fill receipts when the tracker is empty after a restart, and the recovery service now fires the settlement callback in the PostFill branch that previously skipped it. The unconditional callback in `TransactionHandler::handle_confirmed` is safe — `SettlementService::handle_transaction_confirmed` gates on `PostFill | PreClaim` and no-ops for everything else, and error propagation matches the old path. The attestation-data refactor produces byte-identical output (64 ASCII hex bytes), so `can_claim` decoding is unaffected.

Non-blocking notes:
- A missing `DispatchId` event during recovery classifies the order as `Unknown` and will stall it at that stage indefinitely (retried each loop, warn-logged). Acceptable, but worth keeping in mind operationally.
- `gas_payment: U256::ZERO` in the recovery path is a pre-existing TODO that recovered tracker entries will now also hit.
- `transaction_receipt_from_alloy` is a one-line wrapper around `TransactionReceipt::from`; could be inlined.

PR #392, opened 2026-06-10
h02-h06-m24 - Keep transient settlement errors retryable

nahimterrazas:
## Summary

- Preserve typed settlement and delivery errors through post-fill, pre-claim, and claim paths so retryable infrastructure failures are not collapsed into terminal order failures.
- Route settlement-stage failures through a shared transient/permanent policy, leaving retryable orders in their current non-terminal state for recovery to re-drive.
- Persist confirmed PostFill and PreClaim status/hash updates before settlement callback post-processing, and replay missing confirmed callbacks during recovery with a broadcaster-safe gate.
- Add regression coverage for transient settlement errors, confirmed callback failures, recovery callback replay, and recovery re-drive into PreClaimReady.

## Root Cause

Successful on-chain settlement progress could be followed by transient off-chain failures such as RPC, prover, delivery, or callback storage errors. Several engine arms treated those errors as fatal and wrote `OrderStatus::Failed`, which is terminal and excluded from recovery, stranding already-paid fills.

## Verification

- `cargo fmt --all -- --check`
- `cargo test -p solver-core`
- `cargo check --all-targets --all-features`
- `cargo clippy --all-features --all-targets -- -D warnings --allow deprecated`

## Release Notes

* **Bug Fixes**
  * Fixed orders being permanently marked as failed due to transient settlement infrastructure issues; the system now distinguishes retryable errors from permanent failures and preserves order status for transient failures.

* **Improvements**
  * Enhanced recovery mechanism to replay settlement callbacks for orders with confirmed on-chain progress, enabling safer retry of post-fill and pre-claim operations.

* **Tests**
  * Added regression tests validating settlement failure recovery and callback replay for confirmed transactions.


pepebndc (approved):
Reviewed for fund-safety and correctness. The exhaustive `settlement_failure_policy` match (no wildcard arms, compile-error on new upstream variants) is the right design, and the status-advance-before-callback reordering in `handle_confirmed` correctly separates the durable on-chain fact from the retryable off-chain side effect. The PostFill recovery replay gate correctly protects the broadcaster's `proof = None` reset; Hyperlane's `track_submission` is idempotent so the default `recover_post_fill_state → false` is safe there.

Non-blocking notes:
- `SettlementError::Delivery(#[from] DeliveryError)` makes it possible to bypass `map_delivery_error` with a bare `?`, landing `InsufficientNativeGas` in the `Delivery` variant instead of the named one. Both classify as `RetryLater` today so no bug, but worth consolidating eventually.
- `claim_batch.clear()` after `std::mem::take(&mut claim_batch)` is a no-op and can be dropped.
- `Storage`/`State` errors after a successful `deliver()` in `process_claim_batch` still map to `FailOrder` while the claim tx is in flight — pre-existing and explicitly scoped out per the plan doc, but flagging it so the typed post-delivery variant follow-up doesn't get lost.

PR #397, opened 2026-06-18
fix: admin auth hardening (M-18, M-25)

nahimterrazas:
## Summary

Admin-auth hardening for **M-18** and **M-25** (includes review fixes).

### M-18 — Arbitrary spender approvals bypass the withdrawal allowlist
`handle_approve_tokens` forwarded an admin-supplied `spender` straight to on-chain `approve` with no allowlist check, bypassing the `recipient_allowed` control the sibling `handle_withdrawal` enforces — a compromised admin key could `approve(attacker, U256::MAX)` across all tokens/chains. The spender is now validated **per affected network**: it must be a configured input/output settler on *every* chain the approval would touch — the specific requested `chain_id`, or every known network for the all-chains scope (an empty affected set is rejected). A settler configured only on another chain no longer passes (closes the cross-chain address-confusion gap).

### M-25 — Removed admins retain self-renewing API access
- **Refresh re-derives scope from live config:** admin-scoped refresh no longer copies the embedded scope; it re-resolves from the live whitelist (`resolve_siwe_runtime_config` → `resolve_siwe_scopes`). A `sub` no longer whitelisted is rejected (401); a demoted admin's scope is re-derived from the current role.
- **Admin refresh works in admin-only configs:** the `orders_auth_enabled` gate is applied **after** decoding the refresh claims and only to non-admin (public-client) tokens, so a valid admin SIWE refresh is not blocked when public-client auth is disabled.
- **Immediate revocation:** `auth_middleware` rechecks the **live admin whitelist** for admin-scoped access tokens, so a removed admin's still-unexpired token is rejected at once rather than lingering until expiry. Public-client routes are never whitelist-gated.

## Verification
- `cargo test -p solver-service`: 686 passed (incl. per-chain rejection, all-chains-partial rejection, admin-refresh-with-public-auth-off, removed-admin-middleware-rejection)

### DIFF
--- a/crates/solver-service/src/apis/admin.rs
+++ b/crates/solver-service/src/apis/admin.rs
@@ -915,30 +915,34 @@ async fn ensure_new_token_approvals(
 	Ok(approvals_set)
 }
 
-/// Collects the set of configured settler addresses across all networks the
-/// `TokenManager` currently knows about.
+/// Builds a per-network map of configured settler addresses.
 ///
-/// This is the exact set the safe auto-approval path (`ensure_new_token_approvals`)
-/// is willing to grant allowances to: the non-zero input and output settler
-/// addresses on each network. The admin `approve` endpoint validates its
-/// admin-supplied `spender` against this set so a compromised admin key cannot
-/// grant an arbitrary attacker unlimited allowance — the same blast-radius
-/// reduction the withdrawal recipient allowlist provides for `handle_withdrawal`.
-async fn configured_settler_set(
+/// The value for each chain is the set of non-zero input/output settler
+/// addresses the safe auto-approval path (`ensure_new_token_approvals`) is
+/// willing to grant allowances to ON THAT chain. Keeping the mapping
+/// per-network — rather than a global union — is what lets the admin `approve`
+/// endpoint reject cross-chain address confusion: a settler configured on
+/// chain B must not authorize an approval scoped to chain A, where it has no
+/// settler role.
+async fn configured_settlers_by_chain(
 	token_manager: &TokenManager,
-) -> std::collections::HashSet<solver_types::Address> {
+) -> std::collections::HashMap<u64, std::collections::HashSet<solver_types::Address>> {
 	let zero_address = solver_types::Address(vec![0u8; 20]);
 	let networks = token_manager.get_networks().await;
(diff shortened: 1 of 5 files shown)


pepebndc (changes requested):
Reviewed against M-18 and M-25. Both stated findings are correctly closed: per-chain settler validation on `handle_approve_tokens` (M-18), and a live-whitelist recheck in both the `/auth/refresh` path and `auth_middleware`, with scopes re-derived from the current role to cover the demotion case (M-25). Good work; test coverage for the stated claims is solid.

Requesting changes on one security-adjacent gap that falls out of the same M-25 threat model:

**Revoked `AdminAll` token still passes on public routes.** `AuthScope::AdminAll.grants(_)` returns `true` for every scope, including `CreateQuotes` / `CreateOrders` / `ReadOrders`. The public routes `/quotes`, `POST /orders`, and `GET /orders/{id}` are wired with `admin_whitelist_config: None`, so the live-whitelist recheck block in `auth/middleware.rs` is never entered for them. A removed/demoted admin holding a still-valid `AdminAll` access token therefore retains access to those three routes for the access-token lifetime, even though M-25's intent was that revocation take effect immediately everywhere.

Suggested fix: gate the recheck on `is_admin_scoped` alone and fail closed when no whitelist config is available for an admin-scoped token, rather than skipping the recheck when `admin_whitelist_config` is `None`.

Minor (non-blocking): no test for the zero-configured-networks case in settler validation (`approve` should fail closed when no settlers are configured); a small explicit test would document the fail-closed behavior.

pepebndc (approved):
Re-reviewed against the prior CHANGES_REQUESTED items — all addressed.

- M-18/M-25 core findings remain correctly closed.
- Blocking gap (revoked `AdminAll` on public routes) fixed: middleware recheck now gated on `is_admin_scoped` alone and fails closed when no whitelist config is attached; public routes now also receive the live config, so still-whitelisted admins pass and removed admins are rejected immediately. Covered by the fail-closed test and the public-route live-recheck test.
- Non-blocking nit addressed: `test_approve_all_chains_fails_closed_when_no_networks_configured` (+ partial-settler rejection).
- CI fully green.

LGTM 🚀

PR #398, opened 2026-06-18
fix: intake validation hardening (L-01, M-09)

nahimterrazas:
## Summary

Audit hardening for two intake-validation findings (**L-01**, **M-09**), one commit each.

- **L-01 — Unhandled panic on malformed off-chain order.** `from_permit2` and `from_batch_compact` (`crates/solver-types/src/standards/eip7683.rs`) read output `token`/`recipient`/`oracle`/`settler` with bare `.as_str().unwrap()`, reachable pre-signature-check from the unauthenticated `/intent` and `/orders` paths via `StandardOrder::try_from` — a crafted JSON body (missing field or non-string) panicked the request task. The 8 bare unwraps are replaced with `.ok_or("Missing or invalid <field> in output")?` (the functions return `Result<Self, Box<dyn Error>>`, matching the `.ok_or(...)?` style already used a few lines above). Malformed input now returns a clean `Err`.

- **M-09 — Static quote-time gas underprices large callbackData.** callbackData had no byte-length cap at quote/order intake, so size-blind static gas on the origin open/claim legs underpriced large-callback orders. A single shared constant `MAX_CALLBACK_DATA_BYTES = 8192` is introduced in `solver-types` and enforced at three sites that now agree on one number: the existing core guard `validate_callback_calldata` (`cost_profit.rs`, tightened from `u16::MAX`), the quote-side `validate_intent_structure` (`apis/quote/validation.rs`), and the order-side `validate_order` (`solver-order/.../_7683.rs`). 8 KiB comfortably exceeds any legitimate callback while bounding the size-blind calldata gas (the prior 64 KiB `u16::MAX` was far larger than needed).

> Note: the shared `MAX_CALLBACK_DATA_BYTES` declaration lives in `eip7683.rs`, so it rides in the L-01 commit due to file locality; the M-09 commit carries the three enforcement sites.

## Verification

- `cargo test -p solver-types` (298) + the feature-gated `--features oif-interfaces` L-01 panic-safety test, `cargo test -p solver-service` (676), `cargo test -p solver-order` (61): all pass, including the new RED→GREEN tests at each M-09 site and the L-01 malformed-output test.
- `cargo clippy --all-features --all-targets -- -D warnings --allow deprecated`: clean
- `cargo fmt --all -- --check`: clean
- `cargo check --all-targets --all-features`: clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced callback data size limits during order validation and quote requests.
  * Improved gas estimation with optional live fill capability based on configuration.
  * Added stricter validation for order signatures and callback payloads.

* **Tests**
  * Added comprehensive test coverage for callback data size limits and signature validation across order flows.


pepebndc (changes requested):
Reviewed against L-01 and M-09 (diff verified at 4 files / +247 / -14 against the current `main` merge-base).

**L-01: correctly fixed.** The bare `.unwrap()`s in both `from_permit2` and `from_batch_compact` are now `.ok_or(...)?`, so a malformed off-chain order returns a clean `400` instead of panicking. Non-blocking nit: only the `from_permit2` (Permit2/`OifEscrowV0`) path has a malformed-input regression test; the `from_batch_compact` (`OifResourceLockV0`) path is fixed but untested. Please add a malformed-input test for it.

**M-09: only partially addressed.** The 8 KB `callbackData` cap at quote validation (`validation.rs`) and order intake (`_7683.rs`) is the right move and bounds the dominant variable-size term. But measured against the finding, the remediation is incomplete:

1. `enable_live_gas_estimate = false` is still hardcoded at `cost_profit.rs:1746`, so the execution-time profitability gate remains size-blind. The finding explicitly recommends removing this.
2. Only `callbackData` is capped. The finding names `callbackData`, `context`, and signature byte length; `context` and signature are not bounded here. (If `context` is already inherently bounded by `validate_supported_output_context` for every supported type, please confirm that in the PR so it's on record.)

Requesting changes so M-09 isn't marked resolved on the platform while the hardcoded live-estimate flag and the remaining variable-length fields are still open. Either remove the hardcoded flag + add the `context`/signature caps, or scope this PR to "callbackData cap" and track the rest as the M-09 residual.

pepebndc (approved):
Re-reviewed against the prior CHANGES_REQUESTED items — all addressed.

- L-01: `from_batch_compact` malformed-input test added (`test_from_batch_compact_malformed_output_returns_err_not_panic`).
- M-09 fully addressed (beyond the original callbackData-only scope):
  - Hardcoded `enable_live_gas_estimate = false` removed — now reads `config.gas.live_fill_estimate_enabled` and uses the real generated fill tx instead of a synthetic empty one.
  - Signature bounded, lock-type-aware: `MAX_ESCROW_SIGNATURE_BYTES = 66`, `MAX_ORDER_SIGNATURE_BYTES = 8192`.
  - `context` confirmed inherently bounded by `validate_supported_output_context` (empty or 1-byte limit-order only; all else rejected).
  - callbackData 8 KiB cap unified across the three sites.

Approving. Two follow-ups before merge (non-blocking on the review):
1. Please update the PR description — it still scopes M-09 to 'cap callbackData' and doesn't mention the live-estimate flag removal, signature caps, or the context confirmation. Worth reflecting what actually landed so the audit record is accurate.
2. The full Rust CI suite doesn't appear to have run here (only CodeRabbit shows) — please make sure build/test/clippy/e2e go green before merging.

PR #399, opened 2026-06-18
fix: overwrite stale stage hash on confirmed-replacement conflict (M-03)

nahimterrazas:
## Summary

**M-03 — Stale order stage hash strands orders with confirmed same-nonce replacements** (includes review fix).

A same-nonce gas-bump replacement can confirm a *different* tx than the one stored in the order's stage hash, stranding the order at Claim/PreClaim/PostFill/Prepare — recovery resolves to the stale stored hash whose receipt no longer exists, and the order never advances. Two paths fixed:

- **Live duplicate-confirmation** (`crates/solver-core/src/handlers/transaction.rs`): `reconcile_already_applied_stage_hash` now overwrites the stored stage hash with the confirmed (in-hand receipt) hash and still emits `TransactionCanonicalHashConflict`. The unconditional overwrite is sound — the branch runs past the `receipt.success` gate and on-chain idempotency guarantees exactly one confirmation per nonce, so `observed_hash` is the canonical tx (documented in-code).

- **Recovery resolution** (`crates/solver-core/src/recovery/mod.rs`): a new **Layer 0** in `resolve_stage` prefers the authoritative *confirmed-lineage* attempt-ledger hash over a possibly-stale order field for **all** stages (and repairs the field), so a replacement that confirmed while the solver was down is resolved correctly — not just on the Fill stage. A *pending-only* ledger entry never overrides a present order field.

Idempotency preserved: resolution only re-points stored state at a chain-confirmed tx; it triggers no new on-chain action.

## Verification
- `cargo test -p solver-core`: 382 passed (updated `canonical_hash_conflict` tests + new confirmed-lineage and pending-only recovery tests)

### DIFF
--- a/crates/solver-core/src/handlers/transaction.rs
+++ b/crates/solver-core/src/handlers/transaction.rs
@@ -95,12 +95,29 @@ impl TransactionHandler {
 		match stage_hash(order, tx_type) {
 			Some(stored_hash) if stored_hash == observed_hash => Ok(()),
 			Some(stored_hash) => {
+				let stored_hash = stored_hash.clone();
+				// A different confirmed tx landed for this stage (e.g. a same-nonce
+				// gas-bump replacement). The in-hand confirmation receipt proves
+				// `observed_hash` is the canonical tx, so overwrite the stored hash
+				// with it; otherwise recovery would resolve the stage to the
+				// superseded `stored_hash` whose receipt no longer exists and the
+				// order would stall forever. The overwrite always targets a
+				// chain-confirmed tx and produces no new on-chain action, so it is
+				// consistent with the OIF idempotency invariant.
+				let canonical_hash = observed_hash.clone();
+				self.state_machine
+					.update_order_with(order_id, |order| {
+						set_stage_hash(order, tx_type, canonical_hash.clone());
+					})
+					.await
+					.map_err(|e| TransactionError::State(e.to_string()))?;
+
 				self.event_bus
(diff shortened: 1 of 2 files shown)


pepebndc (approved):
Reviewed against the M-03 finding and the surrounding state machine.

The fix is mechanically correct and idempotency-safe. It takes the audit's sanctioned "persist the confirmed replacement hash back onto the stage field" path: `reconcile_already_applied_stage_hash` now overwrites the stale stage hash with the actually-confirmed one (via the CAS-backed `set_transaction_hash` / `update_order_with`), and `resolve_stage` gains a Layer 0 that prefers the confirmed-lineage hash from the attempt ledger. No path through the new code triggers a new on-chain action, so the single fill/claim/settle invariant is preserved. The overwrite is reached only via the `AlreadyApplied` outcome, where the in-hand receipt is proof of canonicity, so the unconditional overwrite is safe. Covers all five stages, not just Fill.

Approving. One non-blocking nit to consider in a follow-up:

- The new Layer-0 tests (`resolve_stage_prefers_confirmed_lineage_over_stale_order_field_non_fill`, `resolve_stage_keeps_order_field_when_ledger_only_pending`) only exercise `PostFill`. The highest-value origin-chain stages from a stranded-funds standpoint, `PreClaim` and `Claim`, have no named test on the Layer-0 path. The logic is generic so coverage exists, but an explicit `Claim`/`PreClaim` case would be a good regression anchor.

PR #400, opened 2026-06-18
fix: delivery-pipeline robustness (M-12, M-13, M-15)

nahimterrazas:
## Summary

Delivery-pipeline robustness for **M-12**, **M-13**, **M-15** (M-13 includes review fix).

- **M-13 — Contended transaction permit stalls the event loop.** The transaction permit is acquired **inside** the spawned task, so a held permit no longer parks the `select!` event loop. Additionally, a **bounded dispatch semaphore** (`MAX_INFLIGHT_HANDLERS = 1024`) is acquired **before** spawn, so a bounded intake (the intent queue / event bus) cannot fan out into unlimited parked waiter tasks. Backpressure engages only at true saturation — far above the single transaction permit — so this fixes both the original event-loop stall and the unbounded-waiter regression.

- **M-12 — Unbounded RPC awaits stall confirmation monitoring past its timeout.** `poll_for_confirmation` wraps each `get_receipt` / `get_block_number` in `tokio::time::timeout(remaining_budget)`, and the HTTP transport is built via `http_with_client` with explicit connect (10s) and request (30s) timeouts, preserving the existing `RetryBackoffLayer`. An endpoint that accepts but never answers can no longer hang past the confirmation deadline.

- **M-15 — Misclassified EIP-1559 priority-fee rejection leaks nonces.** Added the canonical geth `"max priority fee per gas higher than max fee per gas"` (`ErrTipAboveFeeCap`) string to the `DefinitelyRejected` allowlist so the nonce is reclaimed instead of leaked.

## Verification
- `cargo test -p solver-core`: 382 passed (bounded-dispatch + no-stall tests); `cargo test -p solver-delivery`: 146 passed, no hang (probe-hang timeout test + priority-fee classifier test)

## Summary by CodeRabbit

* **Bug Fixes**
  * Added concurrency limits for internal task handling to prevent resource exhaustion during high-activity periods
  * Implemented explicit network timeouts for blockchain confirmation monitoring to improve reliability and prevent indefinite hangs
  * Enhanced transaction error classification for EIP-1559 fee validation failures to prevent nonce state management issues

### DIFF
--- a/crates/solver-core/src/engine/mod.rs
+++ b/crates/solver-core/src/engine/mod.rs
@@ -44,6 +44,22 @@ use tracing::instrument;
 
 const INTENT_QUEUE_CAPACITY: usize = 1024;
 
+/// Upper bound on handler tasks that have been spawned but not yet completed.
+///
+/// `spawn_handler` acquires one of these permits BEFORE `tokio::spawn` and the
+/// spawned task holds it for its whole lifetime. This caps the number of
+/// in-flight (including parked) handler tasks, so a bounded intake — the H-09
+/// intent queue / event bus — cannot be drained into unlimited parked
+/// semaphore-waiter tasks (unbounded memory growth).
+///
+/// Sized to match `INTENT_QUEUE_CAPACITY`: the queue can hold at most that many
+/// pending items, so allowing the same number of concurrent handlers means the
+/// common path never blocks on this bound — backpressure only engages at true
+/// saturation. It is far above the single-permit `transaction_semaphore`, so a
+/// merely contended transaction permit never blocks dispatch (preserving the
+/// M-13 property that a held transaction permit cannot stall the event loop).
+const MAX_INFLIGHT_HANDLERS: usize = INTENT_QUEUE_CAPACITY;
+
 /// Errors that can occur during engine operations.
 ///
 /// These errors represent various failure modes that can occur while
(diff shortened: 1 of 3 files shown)


pepebndc (approved):
Reviewed against M-12, M-13, and M-15. All three are correctly implemented.

- M-13: the new `dispatch_semaphore` (1024 permits) is acquired before `tokio::spawn`, while the inner `transaction_semaphore`/`general_semaphore` are acquired inside the spawned task. The dispatch permit is bound to the task body, so it is released even on a handler panic (tokio unwinds and runs Drop). Sound bounded-backpressure design; the arm body returns promptly instead of parking the select loop.
- M-12: both `get_receipt` and `get_block_number` in `poll_for_confirmation` are wrapped in `tokio::time::timeout(min(remaining, RPC_PROBE_CALL_TIMEOUT), ...)`, and a transport-level reqwest request/connect timeout is set when building the client. Deadline is now enforced.
- M-15: the priority-fee admission string (`max priority fee per gas higher than max fee per gas`) is added to `DefinitelyRejected`, matched case-insensitively, and the nonce is rolled back exactly once. Combined with the base-fee strings already added in #378, M-15 is fully covered.

Approving. One non-blocking nit:

- `spawn_handler_bounds_inflight_tasks` has a vacuous assertion: the `dispatched` counter is incremented only inside the handler body, which never runs while `tx_semaphore` is held for the whole test, so `live <= CAP` is always `0 <= 2` and proves nothing. The meaningful assertion is the existing `!driver.is_finished()`. Consider replacing the counter check with `assert_eq!(dispatch.available_permits(), 0)` after the sleep to directly verify dispatch permits are consumed.

PR #401, opened 2026-06-19
Reject invalid mock pricing amounts

samsamtrum:
Mock pricing parsed input amounts as `f64` and then used the parsed value directly. That allowed negative, `NaN`, or infinite amounts to flow through conversions and produce invalid quotes or wei conversion attempts.

This adds a shared nonnegative finite amount parser for mock pricing conversion inputs and covers both asset conversion and currency to wei paths with regression tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened input validation for currency conversion and pricing by rejecting negative, non-finite (NaN), and infinite numeric amounts with consistent `InvalidData` errors.
  * Same-currency conversions now validate inputs instead of returning unchanged values, with clearer error reporting.

* **Tests**
  * Added/updated tests to ensure `InvalidData` is returned for invalid amount inputs across conversion and pricing-related calculations.


nahimterrazas (approved):
LGTM

PR #413, opened 2026-07-02
feat(pricing): add CoinPaprika pricing implementation

donbagger:
# Summary

Adds a CoinPaprika-backed `PricingInterface` implementation alongside the existing CoinGecko and DefiLlama sources, registered as `"coinpaprika"`.

Why CoinPaprika: the ticker API is free with no API key and no sign-up, which makes it a practical primary or fallback feed for permissionless solver operators - one more independent source to chain in `PricingService` fallbacks. An optional `api_key` switches to the api-pro tier (higher limits) automatically.

Design notes:
- Mirrors `coingecko.rs` one-to-one so the shape is familiar: TTL price cache + short negative cache, in-flight fetch deduplication, schedule-based rate limiter with a bounded sleep cap, request timeout, `custom_prices`, and a `token_id_map` override.
- CoinPaprika ids are `{symbol}-{name}` shaped, so the source carries its own `COINPAPRIKA_TOKEN_MAPPINGS` (ETH -> `eth-ethereum`, USDC -> `usdc-usd-coin`, including a real WETH id) instead of reusing the CoinGecko-shaped `DEFAULT_TOKEN_MAPPINGS`.
- USD-only parity with the existing sources is kept (non-USD requests return `PriceNotAvailable`, same as CoinGecko).
- The ticker endpoint is per-coin (`GET /v1/tickers/{coin_id}?quotes=USD`), which matches how `get_price` already fetches - one id per call.
- Helpers are self-contained in the file rather than factored into shared infra, following the existing convention (`defillama.rs` does the same). Happy to extract shared plumbing in a follow-up if you'd prefer.

I work on developer relations for CoinPaprika, so flagging that interest openly. The implementation only touches the pricing crate and adds no default behavior change - nothing selects it unless configured.

# Testing Process

- `cargo fmt --check`, `cargo clippy -p solver-pricing --all-targets -- -D warnings`, `cargo build -p solver-pricing` all clean
- `cargo test -p solver-pricing`: 122 passed, 0 failed (30 new tests, all offline via wiremock local servers - same-asset conversion, unsupported pairs, custom base_url retrieval, non-USD rejection, unknown token, invalid amounts, cache TTL and negative cache, in-flight dedup, rate-limit sleep cap, timeout bounds, api-pro key routing)
- Verified end-to-end against the live API with the default config before opening this PR: `convert_asset("ETH","USD","1.0")` -> 1642.13, `wei_to_currency("1000000000000000000","USD")` -> 1642.13, `currency_to_wei("3000","USD")` -> 1826895156733445020, and ETH/SOL correctly rejected as non-USD

# Checklist

- [x] Add a reference to related issues in the PR description. (No existing issue - this adds a new pricing source; happy to open one if you track features that way.)
- [x] Add unit tests if applicable.

## Summary by CodeRabbit

* **New Features**
  * Added a new CoinPaprika-based pricing provider using real-time USD quotes.
  * Enabled custom fixed per-token prices and enhanced conversion support (USD-based and wei-to-fiat via ETH/USD).
  * Included caching and concurrent request sharing to reduce redundant upstream requests.
* **Bug Fixes**
  * Improved behavior for unsupported pairs, unavailable conversions, and failure/timeout handling (including safer wait behavior).
* **Tests**
  * Added extensive coverage for configuration validation, caching/deduplication, timeouts, and conversion edge cases.
* **Chores**
  * Registered the new provider so it appears in the available implementations list.


nahimterrazas (approved):
Thanks for your contributions @donbagger!

Approving it.

